PR #5, opened 2020-10-15
`PseudoPotentialFamily`: deduplicate existing pseudos

### LINKED ISSUE #4: Auto-deduplicate pseudo potential nodes upon creation
This feature was omitted intentionally because in `aiida-core v2.0` the new repository implementation will automatically deduplicate the content written to the file repository. However, file content deduplication is not the only argument in favor of pseudo deduplication. If you create two different families that each contain the same pseudo, it may be counter intuitive to have two different nodes, even though they have the same `md5` attribute. It may be more logical to have a single node, that just happens to be part of two families.

sphuber:
Fixes #4 

When creating a new family through the `create_from_folder` class
method, the code will now check for each parsed pseudo potential whether
the exact same potential already exists in the database, and use that
instead. The check is performed on the exact pseudo potential node type
and the md5 of its file content. The behavior can be turned off through
the new `deduplicate` keyword.

The advantage of deduplication is not only a question of efficiency,
where it prevents that the same pseudo potential gets stored as a node
multiple times in the database, but it can also be more intuitive. If
one imports two families that each happen to contain the exact same
pseudo potential, one might expect that there is a single node that
represents that pseudo that just happens to be part of both groups that
represent the two pseudo potential families.

### DIFF
--- a/aiida_pseudo/groups/family/pseudo.py
+++ b/aiida_pseudo/groups/family/pseudo.py
@@ -5,7 +5,7 @@ import re
 from typing import Union, List, Tuple, Mapping
 
 from aiida.common import exceptions
-from aiida.common.lang import type_check
+from aiida.common.lang import classproperty, type_check
 from aiida.orm import Group, QueryBuilder
 from aiida.plugins import DataFactory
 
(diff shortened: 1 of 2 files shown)


sphuber:
> Will there be a CLI for it?

Do you mean whether there will be an option on the install family command to toggle the deduplication?

I think I will merge this and then see if I can add this and the other suggestion

PR #11, opened 2020-10-26
Adding Psp8 (Abinit type) pseudo support. 

sponce24:


aiida_pseudo/data/pseudo/psp8.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
+"""Module for data plugin to represent a pseudo potential in UPF format."""
```
sphuber, line 2:
```suggestion
"""Module for data plugin to represent a pseudo potential in Psp8 format."""
```

aiida_pseudo/data/pseudo/__init__.py
```diff
@@ -5,5 +5,6 @@
 from .psf import *
 from .psml import *
 from .upf import *
+from .psp8 import *
```
sphuber, line 8: Please keep alphabetical order, so move up one line, just before `from .upf import *`

aiida_pseudo/data/pseudo/__init__.py
```diff
@@ -5,5 +5,6 @@
...
 
-__all__ = (pseudo.__all__ + psf.__all__ + psml.__all__ + upf.__all__)
+__all__ = (pseudo.__all__ + psf.__all__ + psml.__all__ + upf.__all__ + psp8.__all__)
```
sphuber, line 10:
```suggestion
__all__ = (pseudo.__all__ + psf.__all__ + psml.__all__ + psp8.__all__ + upf.__all__)
```

aiida_pseudo/data/pseudo/psp8.py
```diff
@@ -0,0 +1,34 @@
...
+from typing import BinaryIO
+
+from .pseudo import PseudoPotentialData
+
+__all__ = ('Psp8Data',)
+
+
+def parse_element(stream: BinaryIO):
+    """Parse the content of the UPF file to determine the element.
```
sphuber, line 11:
```suggestion
    """Parse the content of the Psp8 file to determine the element.
```

aiida_pseudo/data/pseudo/psp8.py
```diff
@@ -0,0 +1,34 @@
...
+
+    :param stream: a filelike object with the binary content of the file.
+    :return: the symbol of the element following the IUPAC naming standard.
+    """
+    lines = stream.read().decode('utf-8')
+
+    return lines.split()[0].strip()
```
sphuber, line 18: Does this line actually return the element symbol? According to the [documentation](https://docs.abinit.org/developers/psp8_info/) the first line of Psp8 is just a title. Shouldn't you be using the atomic number on the second line to determine the element?

aiida_pseudo/groups/family/__init__.py
```diff
@@ -6,5 +6,6 @@
 from .psml import *
 from .sssp import *
 from .upf import *
+from .psp8 import *
 
-__all__ = (pseudo.__all__ + psf.__all__ + psml.__all__ + sssp.__all__ + upf.__all__)
+__all__ = (pseudo.__all__ + psf.__all__ + psml.__all__ + sssp.__all__ + upf.__all__ + psp8.__all__)
```
sphuber, line 11:
```suggestion
__all__ = (pseudo.__all__ + psf.__all__ + psml.__all__ + psp8.__all__ + sssp.__all__ + upf.__all__)
```

aiida_pseudo/groups/family/__init__.py
```diff
@@ -6,5 +6,6 @@
 from .psml import *
 from .sssp import *
 from .upf import *
+from .psp8 import *
```
sphuber, line 9: move two lines up

setup.json
```diff
@@ -21,14 +21,16 @@
             "pseudo = aiida_pseudo.data.pseudo.pseudo:PseudoPotentialData",
             "pseudo.psf = aiida_pseudo.data.pseudo.psf:PsfData",
             "pseudo.psml = aiida_pseudo.data.pseudo.psml:PsmlData",
-            "pseudo.upf = aiida_pseudo.data.pseudo.upf:UpfData"
+            "pseudo.upf = aiida_pseudo.data.pseudo.upf:UpfData",
+            "pseudo.psp8 = aiida_pseudo.data.pseudo.psp8:Psp8Data"
```
sphuber, line 25: please move this one line up to keep alphabetical ordering. Some for the family entry point

sphuber (changes requested):
Thanks @sponce24 . In addition to my comments, please add unit tests for both the data plugin as well as the family. You can literally copy paste those of the `PsfData` for example and just adapt to `Psp8Data`. I think the current implementation is broken, which shows why tests are important

aiida_pseudo/data/pseudo/psp8.py
sponce24: In the case of psp generated with ONCVP(so all the psp from PseudoDojo) the first element of the first line (so the Title line) is the element symbol.

sphuber: But this is not the general rule for all pseudo files that respect the Psp8 format. In that case, I would really not rely on this. The format seems to guarantee the presence of the atomic number though. It is easy to transform this to the symbol in a reliable and reproducible way. Something like the following should work (not tested):
```python
from aiida.common.constants import elements

lines = stream.read().decode('utf-8')

try:
    atomic_number = int(lines.split()[1].strip())
except (IndexError, ValueError):
    raise ValueError('failed to parse the atomic number.')

try:
    symbol = elements[atomic_number]['symbol']
except KeyError:
    raise ValueError(f'the atomic number {atomic_number} is not supported.')

return symbol
```

sphuber:
Another tip, to get the pre-commit running automatically, do the following from the repository:
```
pip install -e .[pre-commit]
pre-commit install
```
This way you will automatically run pre-commit locally and it won't fail remotely

aiida_pseudo/data/pseudo/psp8.py
sponce24: Ok sure.

sphuber:
@sponce24 if you fix the pre-commit, this can be merged

sphuber (approved):

sphuber:
Thanks a lot @sponce24 !

Issue #14, opened 2020-11-01
Add PseudoDojo support with suggested cutoffs


zooks97:
[PseudoDojo](http://www.pseudo-dojo.org)  is a widely-used pseudopotential family developed by the Abinit team based on the ONCVPSP pseudopotential code. Despite being developed by the Abinit team (Abinit uses the psp8 format), PseudoDojo supports a variety of pseudo formats for different codes (Abinit, Q-E, Siesta):

- psp8 : (NC) the pseudo in the psp8 format (abinit)
- upf : (NC) the pseudoo in the upf format (quantum esperesso)
- psml : (NC) the pseudoo in the psml format (siesta)
- xml : (PAW) the paw dataset in xml format (abinit)

[...]

PR #16, opened 2020-11-02
Add support for PseudoDojo

zooks97:
See #14 for comments on PseudoDojo support.

Since my last comment in #14 , I've changed the structure of the PseudoDojo class files: all PseudoDojo families are now in `pseudodojo.py` instead of cluttering the directory with `pseudodojo_[format].py` files.

~~For some reason, the duplicate file test is not passing even though I have not touched the relevant code in `PseudoPotentialFamily`.~~

### DIFF
--- a/aiida_pseudo/cli/__init__.py
+++ b/aiida_pseudo/cli/__init__.py
@@ -7,6 +7,6 @@ import click_completion
 click_completion.init()
 
 from .root import cmd_root
-from .install import cmd_install, cmd_install_family, cmd_install_sssp
+from .install import cmd_install, cmd_install_family, cmd_install_sssp, cmd_install_pseudo_dojo
 from .list import cmd_list
 from .show import cmd_show
--- a/aiida_pseudo/cli/install.py
+++ b/aiida_pseudo/cli/install.py
@@ -248,12 +248,17 @@ def cmd_install_pseudo_dojo(version, functional, relativistic, protocol, pseudo_
                     echo.echo_critical(msg)
 
                 for stringency, element_cutoffs in values['hints'].items():
+<<<<<<< HEAD
 
                     if cutoffs.get(stringency, None) is None:
                         cutoffs[stringency] = {}
 
                     # All supported PseudoDojo potentials are NC, so we take dual of 8.0
                     cutoffs[stringency][element] = {
+=======
+                    # All supported PseudoDojo potentials are NC, so we take dual of 8.0
+                    cutoffs.setdefault(stringency, {})[element] = {
+>>>>>>> Add support for Pseudo Dojo pseudopotential families
                         'cutoff_wfc': element_cutoffs['ecut'],
                         'cutoff_rho': element_cutoffs['ecut'] * 8.0
                     }
--- a/aiida_pseudo/cli/install.py
+++ b/aiida_pseudo/cli/install.py
@@ -6,15 +6,16 @@ import shutil
 import tempfile
 
 import click
-
-from aiida.cmdline.utils import decorators, echo
 from aiida.cmdline.params import options as options_core
 from aiida.cmdline.params import types
+from aiida.cmdline.utils import decorators, echo
 
 from .params import options
 from .root import cmd_root
 
 URL_SSSP_BASE = 'https://legacy-archive.materialscloud.org/file/2018.0001/v4/'
+URL_PSEUDODOJO_BASE = 'http://www.pseudo-dojo.org/pseudos/'
+URL_PSEUDODOJO_METADATA_BASE = 'https://raw.githubusercontent.com/abinit/pseudo_dojo/master/pseudo_dojo/pseudos/'
 
 
 @cmd_root.group('install')
@@ -79,12 +80,11 @@ def cmd_install_sssp(version, functional, protocol, traceback):
     """
     # pylint: disable=too-many-locals
     import requests
-
     from aiida.common.files import md5_file
     from aiida.orm import Group, QueryBuilder
-
     from aiida_pseudo import __version__
     from aiida_pseudo.groups.family import SsspConfiguration, SsspFamily
+
     from .utils import attempt, create_family_from_archive
 
     configuration = SsspConfiguration(version, functional, protocol)
@@ -138,5 +138,126 @@ def cmd_install_sssp(version, functional, protocol, traceback):
 
         family.description = description
         family.set_cutoffs({'normal': cutoffs})
+
+        echo.echo_success(f'installed `{label}` containing {family.count()} pseudo potentials')
+
+
+@cmd_install.command('pseudo-dojo')
+@options.VERSION(type=click.Choice(['03', '04']), default='04')
+@options.FUNCTIONAL(type=click.Choice(['pbe', 'pbesol', 'lda']), default='pbe')
+@options.RELATIVISTIC(type=click.Choice(['sr', 'sr3plus', 'fr']), default='sr')
+@options.PROTOCOL(type=click.Choice(['standard', 'stringent']), default='standard')
+@options.PSEUDO_FORMAT(type=click.Choice(['psp8', 'upf', 'psml']), default='psp8')
+@options.TRACEBACK()
+@decorators.with_dbenv()
+def cmd_install_pseudo_dojo(version, functional, relativistic, protocol, pseudo_format, traceback):
+    """Install a PseudoDojo configuration.
+
+    The PseudoDojo configuration will be automatically downloaded from pseudo-dojo.org to create a new
+    `PseudoDojoFamily` subclass instance based on the specified pseudopotential format.
+    """
+    # pylint: disable=too-many-locals,too-many-arguments,too-many-statements
+    import requests
+    from aiida.common.files import md5_file
+    from aiida.orm import Group, QueryBuilder
+    from aiida_pseudo import __version__
+    from aiida_pseudo.groups.family import (
+        PseudoDojoConfiguration, PseudoDojoPsmlFamily, PseudoDojoPsp8Family, PseudoDojoUpfFamily
+    )
+
+    from .utils import attempt, create_family_from_archive
+
+    cls_mapping = {
+        'psp8': PseudoDojoPsp8Family,
+        'psml': PseudoDojoPsmlFamily,
+        'upf': PseudoDojoUpfFamily,
+    }
+
+    metadata_urls = {
+        'nc-sr-04_pbe_standard': 'ONCVPSP-PBE-PDv0.4/standard.djson',
+        'nc-sr-04_pbe_stringent': 'ONCVPSP-PBE-PDv0.4/stringent.djson',
+        'nc-sr-04_pbesol_standard': 'ONCVPSP-PBEsol-PDv0.4/standard.djson',
+        'nc-sr-04_pbesol_stringent': 'ONCVPSP-PBEsol-PDv0.4/stringent.djson',
+        # 'nc-fr-04_pbe_standard': 'ONCVPSP-PBE-FR-PDv0.4/standard.djson',  # md5 does not match
+        # 'nc-fr-04_pbe_stringent': 'ONCVPSP-PBE-FR-PDv0.4/stringent.djson'  # missing elements
+    }
+
+    try:
+        cls = cls_mapping[pseudo_format]
+    except KeyError:
+        echo.echo_critical(f'{pseudo_format} is not a valid PseudoDojo pseudopotential format')
+
+    configuration = PseudoDojoConfiguration(version, functional, relativistic, protocol)
+    label = cls.format_configuration_label(configuration)
+    description = f'PseudoDojo v{version} {functional} {relativistic} {protocol} {pseudo_format} ' + \
+        f'installed with aiida-pseudo v{__version__}'
+
+    if configuration not in cls.valid_configurations:
+        echo.echo_critical(
+            f'{version} {functional} {relativistic} {protocol}is not a valid PseudoDojo configuration for {cls}'
+        )
+
+    if QueryBuilder().append(cls, filters={'label': label}).first():
+        echo.echo_critical(f'{cls.__name__}<{label}> is already installed')
+
+    with tempfile.TemporaryDirectory() as dirpath:
+
+        if relativistic == 'sr3plus':
+            url_archive = f'{URL_PSEUDODOJO_BASE}/nc-sr-{version}-3plus_{functional}_{protocol}_{pseudo_format}.tgz'
+        else:
+            url_archive = f'{URL_PSEUDODOJO_BASE}/nc-{relativistic}-{version}_{functional}_{protocol}_' + \
+                f'{pseudo_format}.tgz'
+
+        filepath_archive = os.path.join(dirpath, 'archive.tgz')
+
+        with attempt('downloading selected pseudo potentials archive... ', include_traceback=traceback):
+            response = requests.get(url_archive)
+            response.raise_for_status()
+            with open(filepath_archive, 'wb') as handle:
+                handle.write(response.content)
+                handle.flush()
+                description += f'\nArchive pseudos md5: {md5_file(filepath_archive)}'
+
+        with attempt('unpacking archive and parsing pseudos... ', include_traceback=traceback):
+            family = create_family_from_archive(cls, label, filepath_archive)
+
+        family.description = description
+        label_metadata = f'nc-{relativistic}-{version}_{functional}_{protocol}'
+
+        if label_metadata in metadata_urls:
+            url_metadata = f'{URL_PSEUDODOJO_METADATA_BASE}/{metadata_urls[label_metadata]}'
+            filepath_metadata = os.path.join(dirpath, 'metadata.json')
+
+            with attempt('downloading selected pseudo potentials metadata... ', include_traceback=traceback):
+                response = requests.get(url_metadata)
+                response.raise_for_status()
+                with open(filepath_metadata, 'a+b') as handle:
+                    handle.write(response.content)
+                    handle.flush()
+                    handle.seek(0)
+                    metadata = json.load(handle)
+                    description += f'\nPseudo metadata md5: {md5_file(filepath_metadata)}'
+
+            cutoffs = {}
+
+            for element, values in metadata['pseudos_metadata'].items():
+
+                if family.get_pseudo(element).md5 != values['md5']:
+                    Group.objects.delete(family.pk)
+                    msg = f"md5 of pseudo for element {element} does not match that of the metadata {values['md5']}"
+                    echo.echo_critical(msg)
+
+                for stringency, element_cutoffs in values['hints'].items():
+
+                    if cutoffs.get(stringency, None) is None:
+                        cutoffs[stringency] = {}
+
+                    # All supported PseudoDojo potentials are NC, so we take dual of 8.0
+                    cutoffs[stringency][element] = {
+                        'cutoff_wfc': element_cutoffs['ecut'],
+                        'cutoff_rho': element_cutoffs['ecut'] * 8.0
+                    }
+
+            family.set_cutoffs(cutoffs=cutoffs, default_stringency='normal')
 
         echo.echo_success(f'installed `{label}` containing {family.count()} pseudo potentials')
--- a/aiida_pseudo/cli/params/options.py
+++ b/aiida_pseudo/cli/params/options.py
@@ -7,18 +7,40 @@ import click
 from aiida.cmdline.params.options import OverridableOption
 from .types import PseudoPotentialFamilyTypeParam
 
-__all__ = ('VERSION', 'FUNCTIONAL', 'PROTOCOL', 'TRACEBACK', 'FAMILY_TYPE', 'ARCHIVE_FORMAT')
+__all__ = (
+    'VERSION', 'FUNCTIONAL', 'RELATIVISTIC', 'PROTOCOL', 'PSEUDO_FORMAT', 'TRACEBACK', 'FAMILY_TYPE', 'ARCHIVE_FORMAT'
+)
 
 VERSION = OverridableOption(
-    '-v', '--version', type=click.STRING, required=False, help='Select the version of the SSSP configuration.'
+    '-v', '--version', type=click.STRING, required=False, help='Select the version of the installed configuration.'
 )
 
 FUNCTIONAL = OverridableOption(
-    '-f', '--functional', type=click.STRING, required=False, help='Select the functional of the SSSP configuration.'
+    '-x',
+    '--functional',
+    type=click.STRING,
+    required=False,
+    help='Select the functional of the installed configuration.'
+)
+
+RELATIVISTIC = OverridableOption(
+    '-r',
+    '--relativistic',
+    type=click.STRING,
+    required=False,
+    help='Select the type of relativistic effects included in the installed configuration.'
 )
 
 PROTOCOL = OverridableOption(
-    '-p', '--protocol', type=click.STRING, required=False, help='Select the protocol of the SSSP configuration.'
+    '-p', '--protocol', type=click.STRING, required=False, help='Select the protocol of the installed configuration.'
+)
+
+PSEUDO_FORMAT = OverridableOption(
+    '-f',
+    '--pseudo-format',
+    type=click.STRING,
+    required=True,
+    help='Select the pseudopotential file format of the installed configuration.'
 )
 
 TRACEBACK = OverridableOption(
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
 
         strategy:
             matrix:
-                python-version: [3.6, 3.7, 3.8]
+                python-version: [3.6, 3.7, 3.8, 3.9]
 
         services:
             rabbitmq:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Change log
 
+## 0.3.0
+
+### Features
+- Add support for Python 3.9 [[#21]](https://github.com/aiidateam/aiida-pseudo/pull/21)
+- `RecommendedCutoffMixin`: add concept of stringency levels [[#20]](https://github.com/aiidateam/aiida-pseudo/pull/20)
+- `UpfData`: automatically parse the Z valence from the file [[#24]](https://github.com/aiidateam/aiida-pseudo/pull/24)
+
+### Fixes
+- `PseudoPotentialFamily.get_pseudos`: maintain structure kind names [[#20]](https://github.com/aiidateam/aiida-pseudo/pull/20)
+
+
 ## 0.2.0
 
 ### Changes
--- a/aiida_pseudo/__init__.py
+++ b/aiida_pseudo/__init__.py
@@ -1,3 +1,3 @@
 # -*- coding: utf-8 -*-
 """AiiDA plugin that simplifies working with pseudo potentials."""
-__version__ = '0.2.0'
+__version__ = '0.3.0'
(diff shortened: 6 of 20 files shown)


aiida_pseudo/cli/install.py
```diff
@@ -140,3 +140,109 @@ def cmd_install_sssp(version, functional, protocol, traceback):
         family.set_cutoffs(cutoffs)
 
         echo.echo_success(f'installed `{label}` containing {family.count()} pseudo potentials')
+
+
+@cmd_install.command('pseudodojo')
+@options.VERSION(type=click.Choice(['03', '04']), default='04')
+@options.FUNCTIONAL(type=click.Choice(['pbe', 'pbesol', 'lda']), default='pbe')
+@options.REL(type=click.Choice(['sr', 'sr3plus', 'fr']), default='sr')
```
sponce24, line 148: Will this work if the version is 03? There should be only sr in that case I think.

aiida_pseudo/cli/install.py
```diff
@@ -140,3 +140,109 @@ def cmd_install_sssp(version, functional, protocol, traceback):
...
+    from .utils import attempt, create_family_from_archive
+
+    pseudo_format_families = {'psp8': PseudoDojoPsp8Family, 'psml': PseudoDojoPsmlFamily, 'upf': PseudoDojoUpfFamily}
+
+    metadata_urls = {
+        'nc-sr-04_pbe_standard': 'ONCVPSP-PBE-PDv0.4/standard.djson',
+        'nc-sr-04_pbe_stringent': 'ONCVPSP-PBE-PDv0.4/stringent.djson',
+        'nc-sr-04_pbesol_standard': 'ONCVPSP-PBEsol-PDv0.4/standard.djson',
+        'nc-sr-04_pbesol_stringent': 'ONCVPSP-PBEsol-PDv0.4/stringent.djson',
+        # 'nc-fr-04_pbe_standard': 'ONCVPSP-PBE-FR-PDv0.4/standard.djson',  # md5 does not match
```
sponce24, line 178: I think in the case of FR, the hint table has to be taken from the GitHub: https://raw.githubusercontent.com/PseudoDojo/ONCVPSP-PBE-FR-PDv0.4/master/standard.djson

aiida_pseudo/cli/install.py
```diff
@@ -140,3 +140,109 @@ def cmd_install_sssp(version, functional, protocol, traceback):
...
+        # 'nc-fr-04_pbe_stringent': 'ONCVPSP-PBE-FR-PDv0.4/stringent.djson'  # missing elements
```
sponce24, line 179: Same https://raw.githubusercontent.com/PseudoDojo/ONCVPSP-PBE-FR-PDv0.4/master/stringent.djson

aiida_pseudo/cli/params/options.py
```diff
@@ -7,18 +7,30 @@
 from aiida.cmdline.params.options import OverridableOption
 from .types import PseudoPotentialFamilyTypeParam
 
-__all__ = ('VERSION', 'FUNCTIONAL', 'PROTOCOL', 'TRACEBACK', 'FAMILY_TYPE', 'ARCHIVE_FORMAT')
+__all__ = ('VERSION', 'FUNCTIONAL', 'REL', 'PROTOCOL', 'HINT', 'PSEUDO_FORMAT', 'TRACEBACK', 'FAMILY_TYPE', 'ARCHIVE_FORMAT')
 
 VERSION = OverridableOption(
-    '-v', '--version', type=click.STRING, required=False, help='Select the version of the SSSP configuration.'
+    '-v', '--version', type=click.STRING, required=False, help='Select the version of the SSSP/PseudoDojo configuration.'
```
sponce24: Select the version of the SSSP or PseudoDojo

aiida_pseudo/cli/params/options.py
```diff
@@ -7,18 +7,30 @@
...
+    '-x', '--functional', type=click.STRING, required=False, help='Select the functional of the SSSP/PseudoDojo configuration.'
+)
+
+REL = OverridableOption(
+    '-r', '--rel', type=click.STRING, required=False, help='Select the type of relativistic effects included in the PseudoDojo configuration.'
 )
 
 PROTOCOL = OverridableOption(
-    '-p', '--protocol', type=click.STRING, required=False, help='Select the protocol of the SSSP configuration.'
+    '-p', '--protocol', type=click.STRING, required=False, help='Select the protocol of the SSSP/PseudoDojo configuration.'
```
sponce24: Select the version of the SSSP or PseudoDojo

aiida_pseudo/cli/install.py
zooks97: Yes, if you try to give an invalid combination of version / functional / rel / etc, you'll be given an error that it's an invalid configuration.

aiida_pseudo/cli/install.py
zooks97: The md5 hash of the pseudopotential file and the md5 hash listed in the file that you linked don't match. I need to investigate further to find out why and double check what the problem is.

aiida_pseudo/cli/install.py
zooks97: For this one, there are elements which are missing cutoff hints in the json file, so unless we remove those elements from the pseudopotential library, we can't use these cutoffs.

aiida_pseudo/cli/params/options.py
zooks97: Makes sense 👍

aiida_pseudo/cli/__init__.py
```diff
@@ -7,6 +7,6 @@
 click_completion.init()
 
 from .root import cmd_root
-from .install import cmd_install, cmd_install_family, cmd_install_sssp
+from .install import cmd_install, cmd_install_family, cmd_install_sssp, cmd_install_pseudodojo
```
sphuber, line 10: Given that pseudo dojo seems to be two words, and you use `PseudoDojo` as capitalization, it would also follow that we use `pseudo_dojo` in filenames

aiida_pseudo/cli/install.py
```diff
@@ -140,3 +140,111 @@ def cmd_install_sssp(version, functional, protocol, traceback):
         family.set_cutoffs(cutoffs)
 
         echo.echo_success(f'installed `{label}` containing {family.count()} pseudo potentials')
+
+
+@cmd_install.command('pseudodojo')
```
sphuber, line 145:
```suggestion
@cmd_install.command('pseudo-dojo')
```

aiida_pseudo/cli/install.py
```diff
@@ -140,3 +140,111 @@ def cmd_install_sssp(version, functional, protocol, traceback):
...
+@options.VERSION(type=click.Choice(['03', '04']), default='04')
+@options.FUNCTIONAL(type=click.Choice(['pbe', 'pbesol', 'lda']), default='pbe')
+@options.REL(type=click.Choice(['sr', 'sr3plus', 'fr']), default='sr')
+@options.PROTOCOL(type=click.Choice(['standard', 'stringent']), default='standard')
+@options.HINT(type=click.Choice(['low', 'normal', 'high', 'none']), default='none')
+@options.PSEUDO_FORMAT(type=click.Choice(['psp8', 'upf', 'psml']), default='psp8')
+@options.TRACEBACK()
+@decorators.with_dbenv()
+def cmd_install_pseudodojo(version, functional, rel, protocol, hint, pseudo_format, traceback):
+    """Install an PseudoDojo configuration.
```
sphuber, line 155:
```suggestion
    """Install a PseudoDojo configuration.
```

aiida_pseudo/cli/install.py
```diff
@@ -140,3 +140,111 @@ def cmd_install_sssp(version, functional, protocol, traceback):
...
+
+    The PseudoDojo configuration will be automatically downloaded from pseudo-dojo.org to create a new
+    `PseudoDojoFamily` subclass based on the correct pseudopotential format.
```
sphuber, line 158:
```suggestion
    `PseudoDojoFamily` instance based on the correct pseudopotential format.
```

aiida_pseudo/cli/install.py
```diff
@@ -140,3 +140,111 @@ def cmd_install_sssp(version, functional, protocol, traceback):
...
+    from .utils import attempt, create_family_from_archive
+
+    pseudo_format_families = {'psp8': PseudoDojoPsp8Family, 'psml': PseudoDojoPsmlFamily, 'upf': PseudoDojoUpfFamily}
+
+    metadata_urls = {
+        'nc-sr-04_pbe_standard': 'ONCVPSP-PBE-PDv0.4/standard.djson',
+        'nc-sr-04_pbe_stringent': 'ONCVPSP-PBE-PDv0.4/stringent.djson',
+        'nc-sr-04_pbesol_standard': 'ONCVPSP-PBEsol-PDv0.4/standard.djson',
+        'nc-sr-04_pbesol_stringent': 'ONCVPSP-PBEsol-PDv0.4/stringent.djson',
+        # 'nc-fr-04_pbe_standard': 'ONCVPSP-PBE-FR-PDv0.4/standard.djson',  # md5 does not match
```
sphuber, line 178: What will happen if this config is chosen by a user. Will the code except? I think on 221 it will raise a `KeyError`.

aiida_pseudo/cli/params/options.py
```diff
@@ -7,18 +7,56 @@
...
 FUNCTIONAL = OverridableOption(
-    '-f', '--functional', type=click.STRING, required=False, help='Select the functional of the SSSP configuration.'
+    '-x',
+    '--functional',
+    type=click.STRING,
+    required=False,
+    help='Select the functional of the SSSP or PseudoDojo configuration.'
+)
+
+REL = OverridableOption(
```
sphuber, line 30: I'd really prefer this to be fully spelled out. Succinct just makes it less clear
```suggestion
RELATIVISTIC = OverridableOption(
```

aiida_pseudo/groups/family/pseudodojo.py
```diff
@@ -0,0 +1,227 @@
...
+
+    The `PseudoDojoPsp8Family` is essentially a `PseudoDojoFamily` with some additional constraints.
+    It can only be used to contain the pseudo potentials and corresponding metadata of an official psp8
+    PseudoDojo configuration.
+    """
+
+    default_configuration = PseudoDojoConfiguration('04', 'pbe', 'sr', 'standard', 'normal', 'psp8')
+    valid_configurations = (
+        # nc-sr-03_pbe
+        PseudoDojoConfiguration('03', 'pbe', 'sr', 'standard', 'none', 'psp8'),
```
sphuber, line 70: Should `'none'` not just be `None`? Or does the string `'none'` have a special meaning for Pseudo Dojo?

aiida_pseudo/groups/family/pseudodojo.py
```diff
@@ -0,0 +1,227 @@
...
+        PseudoDojoConfiguration('04', 'pbesol', 'sr', 'stringent', 'high', 'psp8'),
+        PseudoDojoConfiguration('04', 'pbesol', 'sr', 'stringent', 'none', 'psp8'),
+        # nc-sr-04_lda
+        PseudoDojoConfiguration('04', 'lda', 'sr', 'standard', 'none', 'psp8'),
+        PseudoDojoConfiguration('04', 'lda', 'sr', 'stringent', 'none', 'psp8'),
+        # nc-sr-04-3plus_pbe
+        PseudoDojoConfiguration('04', 'pbe', 'sr3plus', 'standard', 'none', 'psp8'),
+        # nc-fr-04_pbe_standard
+        # PseudoDojoConfiguration('04', 'pbe', 'fr','standard', 'low', 'psp8'),
+        # PseudoDojoConfiguration('04', 'pbe', 'fr','standard', 'normal', 'psp8'),
```
sphuber, line 105: Are all these configurations really different for each family simply based on the format of the underlying pseudo? The format is merely a way of encoding the definition of the pseudo is it not? So then it is technically possible to take a pseudo in one format and transform it to any other format. If that is so, then the same configurations could exist for each of the three formats, could they not?

aiida_pseudo/groups/family/pseudodojo.py
```diff
@@ -0,0 +1,227 @@
...
+        # PseudoDojoConfiguration('04', 'pbe', 'fr','standard', 'high', 'psp8'),
+        PseudoDojoConfiguration('04', 'pbe', 'fr', 'standard', 'none', 'psp8'),
+        # nc-fr-04_pbe_stringent
+        # PseudoDojoConfiguration('04', 'pbe', 'fr', 'stringent', 'low', 'psp8'),
+        # PseudoDojoConfiguration('04', 'pbe', 'fr', 'stringent', 'normal', 'psp8'),
+        # PseudoDojoConfiguration('04', 'pbe', 'fr', 'stringent', 'high', 'psp8'),
```
sphuber, line 111: What does it mean for these to be commented out?

aiida_pseudo/groups/family/pseudo.py
```diff
@@ -150,6 +151,85 @@ def create_from_folder(cls, dirpath, label, description='', deduplicate=True):
 
         return family
 
+    @classmethod
+    def validate_cutoffs(cls, elements: set, cutoffs: dict) -> None:
```
sphuber, line 155: I originally explicitly did not implement this on the base class, because I was not sure that all pseudo types would 1) need recommended cutoffs like this (the concept might not even make sense) and 2) might require a different implementation/definition. Now that there is a second one, the pseudo dojo pseudos, that seem to be able to use the same concept, I agree that it may seem logical to move this here, but this may not always be the case. Of course then we could create another subclass that implements this, but this may cause trouble with data migrations. Maybe we should already consider doing this now?

aiida_pseudo/cli/install.py
```diff
@@ -140,3 +140,111 @@ def cmd_install_sssp(version, functional, protocol, traceback):
...
 
         echo.echo_success(f'installed `{label}` containing {family.count()} pseudo potentials')
+
+
+@cmd_install.command('pseudodojo')
+@options.VERSION(type=click.Choice(['03', '04']), default='04')
+@options.FUNCTIONAL(type=click.Choice(['pbe', 'pbesol', 'lda']), default='pbe')
+@options.REL(type=click.Choice(['sr', 'sr3plus', 'fr']), default='sr')
+@options.PROTOCOL(type=click.Choice(['standard', 'stringent']), default='standard')
+@options.HINT(type=click.Choice(['low', 'normal', 'high', 'none']), default='none')
```
sphuber, line 150: What is the difference between the protocol and hint? Why does the protocol not also determine what cutoffs should be used with it?

aiida_pseudo/cli/install.py
```diff
@@ -140,3 +140,111 @@ def cmd_install_sssp(version, functional, protocol, traceback):
...
+                handle.write(response.content)
+                handle.flush()
+                description += f'\nArchive pseudos md5: {md5_file(filepath_archive)}'
+
+        with attempt('unpacking archive and parsing pseudos... ', include_traceback=traceback):
+            family = create_family_from_archive(pseudo_dojo_format_family, label, filepath_archive)
+
+        family.description = description
+
+        if hint != 'none':
```
sphuber, line 219: Seeing this concept of the pseudo dojo that there are multiple recommended cutoffs for a family, makes me think whether we should not just include this into the code as a default, i.e., instead of creating multiple groups of a pseudo family, each one with a particular set of recommended cutoffs, maybe we create a single group and add *all* recommended cutoffs and set one of those as the default. I think this makes more sense. The SSSP can just provide the single set as it does now. What do you think? If you agree, I will first make that adaptation to the code in general and then you can just use it for this PR

sphuber (changes requested):
Thanks a lot @zooks97 . It would be great to add support for PD. Have some comments and suggestions. Once those are addressed, we should also definitely add tests

aiida_pseudo/cli/install.py
zooks97: No, the corresponding PseudoDojoConfigurations are also commented, so you will get the `ValueError` in line 54 of `pseudodojo.py` (`pseudo_dojo.py` after I change the naming).

aiida_pseudo/groups/family/pseudodojo.py
zooks97: They can, yes. I did this because the name of the family is generated directly from the `PseudoDojoConfiguration`, and I didn't want to have conflicting names when the same configuration with a different file type is present, which would make it inconvenient to load the correct family by label.

aiida_pseudo/groups/family/pseudodojo.py
zooks97: They are commented out because the suggested cutoff `djson` files are bugged with either missing elements or md5 hashes that do not match the files downloaded from the PseudoDojo website.

aiida_pseudo/groups/family/pseudo.py
zooks97: I did this just because it was the shortest path to me, and I didn't really know what I was doing. It definitely makes sense to implement the cutoff suggestions in a subclass and then go from there if we expect some types of pseudos to never need this feature.

aiida_pseudo/cli/install.py
zooks97: The protocol corresponds to parameters used by the code that generates the pseudopotentials (I think it's GBRV or something). PseudoDojo makes a distinction between this (at the pseudopotential generation level) and the hints, which they determine based on different convergence criteria.

As far as I can tell, the protocol -> a set of pseudopotentials, while the hints -> a set of convergence criteria applied to the set of pseudopotentials which come from the protocol.

aiida_pseudo/cli/install.py
zooks97: I think that would cut down on a lot of repetition in `aiida-pseudo`, and it means in the end, you don't end up with 50 different pseudo families. 

But it kind of kicks the problem down the road. This means that the user then has to pull this information out manually each time they use the pseudo family, and they have to remember what every pseudo family contains in terms of cutoffs (is it none, one set, many sets? what are the names that this specific family calls their different levels of cutoffs?). 

I'm not entirely certain where we want to deal with it. If we do it here, then the cutoffs are uniquely determined by the pseudopotential family from which the pseudos are retrieved, and the level of precision is clearly stated in the pseudo family label. If we do it down the line, then we risk confusing and frustrating users when they have to deal themselves with knowing what kind(s) of cutoff suggestions each family contains and how to properly specify their choice. 

I guess what I'm saying is that in my mind, it's almost a choice between outward cleanliness and ease of use, but that could totally be remedied with some smart implementation details.

aiida_pseudo/cli/__init__.py
zooks97: This makes sense. I'll go through and change it.

aiida_pseudo/groups/family/pseudodojo.py
zooks97: `None` is fine, I wasn't sure if I should keep all entries in the named tuple as strings.

aiida_pseudo/cli/install.py
sponce24: Let me know if you can fix this. Otherwise I will contact the Matteo to see if he can fix the problem.

aiida_pseudo/cli/install.py
sponce24: I think it is important that we support a FR table. Indeed without them we cannot do SOC. Why not using a large value for the ecut for which we do not have the value provided ? I would either use the largest value among all pseudo or something like 60 Ha.

aiida_pseudo/groups/family/pseudodojo.py
sponce24: They are multiple ways to solve this: using a high ecut like I mentioned above or using the corresponding SR hint values. Indeed I believe the FR psp are generated with the same values as the SR one but just solving the Dirac equation instead of Schrodinger. Therefore one would expect the same or close values for energy cutoff.

aiida_pseudo/cli/install.py
sponce24: This is correct but I can elaborate a bit on this. Basically 'standard' psp are generated with the goal of doing DFT or DFPT calculations. For each psp, a set of energy ecut is provided if you want low, normal or high accuracy. 
In the case of many body (like the GW method), you need pseudopotential with more semicore states and more stringent parameters (the cut-off radii beyond which the psp is equal to the all-electron wavefunctions is shorter). 
In this case they generated the 'stringent' set of psp. Then again for that set you have suggested low, normal or high accuracy ecut. 
Hope this clarifies it.

sphuber: I chose the word protocol, but it is not really a term the SSSP paper uses as well. They don't have a term, but I needed one for the plugin. I chose it to distinguish between the two versions they have, the "efficiency" and "precision". Each of these families come with a different set of pseudos (some overlap exists) and an individual set of recommended cutoffs for each set of pseudos. I think the term protocol makes sense, also in the sense we are using it of the `aiida-common-workflows` project. The values of PD `stringent` and `standard` also nicely fall onto that definition, so I think it is fine to keep it. Regarding "hint", I think that term is too general and  vague. Given that in both cases that this applies, both PD as well as the SSSP, it solely is considered the recommended cutoffs, maybe we should be explicit and call it "stringency" and so make the reusable command line option `options.CUTOFF_STRINGENCY('--cutoff-stringency')`. I think that is clear

aiida_pseudo/cli/install.py
sphuber: Fair point, but as I suggested we should of course include a default suggested cutoff set. So even with this change, doing `family.get_recommended_cutoffs(structure=structure)` will still work. However, it will also allow you to do `family.get_recommended_cutoffs(structure=structure, stringency='high')`. We can easily add a method that will provide the available stringencies: `family.get_available_cutoff_stringencies() == ['low', 'normal', 'high']` and `family.get_default_cutoff_stringency() == 'normal'`

zooks97: Please take a look at my implementation

aiida_pseudo/groups/family/pseudodojo.py
zooks97: I don't think I really answered your question the first time.

In theory, one could create files in all the formats with identical pseudopotentials, in which case you would only need one configuration which would apply to all formats.

However, like I said above, some configurations are _currently_ only available for certain subsets of file formats, so I do this to make sure that all the configurations are actually valid.

aiida_pseudo/groups/family/pseudodojo.py
sphuber: would it be possible to get the djson files fixed on the pseudo dojo website @sponce24 ? I imagine this would be in their interest regardless of this plugin.

aiida_pseudo/groups/family/pseudodojo.py
sphuber: > They can, yes. I did this because the name of the family is generated directly from the PseudoDojoConfiguration, and I didn't want to have conflicting names when the same configuration with a different file type is present, which would make it inconvenient to load the correct family by label.

I see, but this can be solved in a more elegant way. You could retrieve it directly from the pseudo type of the family:
```
    @classmethod
    def format_configuration_label(cls, configuration: PseudoDojoConfiguration) -> str:
        """Format a label for an `PseudoDojoFamily` with the required syntax.

        :param configuration: the PseudoDojo configuration
        :return: label
        """
        from aiida.plugins.entry_point import get_entry_point_from_class

        pseudo_type = cls._pseudo_type
        entry_point_group, entry_point = get_entry_point_from_class(pseudo_type.__module__, pseudo_type.__name__)
        pseudo_format = entry_point.name.split('.')[-1]

        return cls.label_template.format(
            version=configuration.version,
            functional=configuration.functional,
            relativistic=configuration.relativistic,
            protocol=configuration.protocol,
            format=pseudo_format
        )
```

Given that the format does not affect the cutoffs, this also makes a lot more sense semantically speaking. This way you can get rid of all the duplication of specifying the same valid configurations over and over again.

The only difference I can see is that version 0.3 is not available for the Psml families. It would be ideal if those could also be generated of course. If this is really not possible, I would instead just invert the logic a bit and define an `unavailable_configuration` just for the `PseudoDojoPsmlFamily` family.

aiida_pseudo/groups/family/pseudodojo.py
zooks97: We're working with the maintainers to get this worked out; it may take some time.

aiida_pseudo/groups/family/pseudodojo.py
zooks97: Seems like a reasonable solution. At the beginning, I was hoping that we could keep everything as similar as possible with the SSSP, but at this point that's clearly out the window.

So, file format will be used by the CLI to select the appropriate `PseudoDojoFamily` subclass, and the labelling will be handled by the subclass?

sphuber: > Seems like a reasonable solution. At the beginning, I was hoping that we could keep everything as similar as possible with the SSSP, but at this point that's clearly out the window.

Well this is a minor difference, and only really in the implementation of `get_valid_labels`. The interface remains the same. You can have a look at the commit I pushed.

> So, file format will be used by the CLI to select the appropriate `PseudoDojoFamily` subclass, and the labelling will be handled by the subclass?

The label formatting is still implemented by the `PseudoDojoFamily` class. The CLI will still call this, but by calling it on one of the subclasses, the format part of the label will be automatically set.

aiida_pseudo/cli/install.py
sphuber: @zooks97 the problem should be in [this file on their repo](https://github.com/abinit/pseudo_dojo/blob/master/pseudo_dojo/pseudos/ONCVPSP-PBE-FR-PDv0.4/standard.djson) then correct? If we are sure we are generating the correct md5, can we simply open a PR that corrects the faulty md5?

zooks97: The md5s in their repo are not faulty, they actually refer to slightly different pseudos than what are on the pseudo-dojo site (and the cutoffs in the file also refer to those tweaked pseudos). Not sure the best way to move forward.

sphuber:
@zooks97 your merge wasn't quite correct. Do you know how to sort it out or you want me to do it?

zooks97:
> @zooks97 your merge wasn't quite correct. Do you know how to sort it out or you want me to do it?

Sorry, should have read this before I nuked the fork.

zooks97:
There are many PseudoDojo archives which have nested folders:
```
nc-fr-04_pbe_standard_psml/:
nc-fr-04_pbe_standard

nc-fr-04_pbe_standard_psp8/:
nc-fr-04_pbe_standard

nc-fr-04_pbe_standard_upf/:
nc-fr-04_pbe_standard

nc-sr_pbesol_standard_psp8/:
pbesol_s_sr

nc-sr_pbesol_standard_upf/:
pbesol_s_sr

nc-sr_pbesol_stringent_psp8/:
pbesol_h_sr

nc-sr_pbesol_stringent_upf/:
pbesol_h_sr

nc-sr_pbe_standard_psp8/:
pbe_s_sr

nc-sr_pbe_standard_upf/:
pbe_s_sr

nc-sr_pbe_stringent_psp8/:
pbe_h_sr

nc-sr_pbe_stringent_upf/:
pbe_h_sr

nc-sr_pw_standard_psp8/:
pw_s_sr

nc-sr_pw_standard_upf/:
pw_s_sr

nc-sr_pw_stringent_psp8/:
pw_h_sr

nc-sr_pw_stringent_upf/:
pw_h_sr

paw_pbe_standard_xml/:
paw_pbe_standard

paw_pbe_stringent_xml/:
paw_pbe_stringent

paw_pw_standard_xml/:
paw_pw_standard

paw_pw_stringent_xml/:
paw_pw_stringent
```

sphuber:
Superseded by #26

sphuber:
Fixed in PR #26

PR #20, opened 2020-11-08
`PseudoPotentialFamily.get_pseudos`: maintain structure kind names

sphuber:
Fixes #91

The method was incorrectly returning a mapping of the pseudos onto the
symbols of the structure's kind, thereby discarding the kind name, which
can often be different from the symbol. For example, `aiida-quantumespresso`
uses kind names that suffix the element symbol with a digit in order to
specify different magnetic kinds.

### DIFF
--- a/aiida_pseudo/groups/family/pseudo.py
+++ b/aiida_pseudo/groups/family/pseudo.py
@@ -254,6 +254,6 @@ class PseudoPotentialFamily(Group):
             raise ValueError('structure should be a `StructureData` instance.')
 
         if structure is not None:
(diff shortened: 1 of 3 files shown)


zooks97 (approved):
Looks all good. In the future, we may need to consider supporting symbols other than [0-9] in the kind names depending on whether codes support doing so. Otherwise, it should be clearly documented somewhere what we support as kind names.

sphuber:
The code actually does not make any assumptions or put any limitations as to what kind names are used. It simply uses whatever kind names the structure contains. It is just in the tests that I use only digits

PR #26, opened 2020-12-07
Add the `PseudoDojoFamily` functionality

zooks97:


zooks97:
URL code is a bit of a joke now, but it gets the job done.

aiida_pseudo/groups/family/pseudo_dojo.py
```diff
@@ -107,6 +306,18 @@ def format_configuration_label(cls, configuration: PseudoDojoConfiguration) -> s
...
         )
 
+    @classmethod
+    def get_urls(cls, label):
+        """Get the pseudo archive and djrepo archive urls for a given configuration label.
+
+        :param label: PseudoDojoConfiguration label
+        :return: dictionary with keys 'pseudo' for pseudo archive and 'djrepo' for djrepo archive urls
+        """
+        urls = cls.urls.get(label)
```
sphuber, line 316: This will just except if `label` is not in the `urls` dictionary won't it? Should probably be `urls = cls.urls.get(label, None)`. But in that case it is better to change this to try/except

aiida_pseudo/groups/family/pseudo_dojo.py
```diff
@@ -0,0 +1,489 @@
...
+        :param djrepo: dictionary loaded from DJREPO JSON file
+        :reutnrs: md5 string
+        """
+        md5_key_mapping = {UpfData: 'md5_upf', Psp8Data: 'md5', PsmlData: 'md5_psml', JthXmlData: 'md5'}
+
+        md5_key = md5_key_mapping.get(pseudo_type)
+        if md5_key is None:
+            raise ValueError(f'pseudo type `{pseudo_type}` is unsupported by PseudoDojo djrepos.')
+        md5 = djrepo.get(md5_key)
+        if md5 is None:
```
sphuber, line 334: same here as for the urls. These should be `get(key, None)` or even better just do try/except the KeyError

sphuber (approved):
Alright, let's get this show on the road! Thanks a lot for the great work @zooks97 . I just realized that it would be nice to get a list of valid configurations from the command line, because now when you ask for an invalid one, it just says that but doesn't give a lot of info. But we keep that for a future feature.

PR #31, opened 2020-12-07
Remove pseudo potential family plugins for specific file formats

### LINKED ISSUE #25: Make the format of pseudo potentials used for a family just an attribute
Currently a pseudo potential family class is made for each pseudo potential format. However, this forces us to create a new entry point for each format times the number pseudo family types, such as SSSP and Pseudo Dojo. Really the format of the pseudo should not matter because it is just the encoding. In principle a pseudo should be able to be transformed into any format without losing information. The format is just important in the sense that certain codes only support certain formats. There is no need to couple the family implementation so closely to it. Instead, we should make the format of pseudos that is used by a certain family just an attribute. This way, one can still query for a family with a certain format.

sphuber:
Fixes #25


zooks97:
Why not support PK and UUID in `load_pseudo_family`? Supporting this here does not preclude loading the `Group` directly but adds the benefit of `load_pseudo_family` being a one-stop-shop for users, no matter what identifier they have on hand.

sphuber:
@zooks97 following the discussion of last week, I have updated the code to decouple the family class from the pseudo types it can host. Any family can now support one or many pseudo types, even though a single instance can only host pseudos of a single type. I added the `load_pseudo_profile` to make loading a specific family by label and a specific pseudo type easier, as now there no longer is a specific class that can serve as a shortcut. Could you have a look if you see any potential problems in the design/implementation and whether this would simplify the implementation of the Pseudo Dojo work?

zooks97:
In the same vein, it could be possible to make `pseudo_type` an optional argument if the user doesn't explicitly know the `pseudo_type` (e.g. installing SSSP, you don't specify a type because there is only UPF, so where are they supposed to get that information?) or knows that only one family exists with the given label. If there is only one family which contains only one type, return that; otherwise, raise a similar exception to the case where multiple families have the same label and type.

zooks97:
> implementation of the Pseudo Dojo work?

Overall, it looks very good. It will simplify the PseudoDojo work as far as reducing the number of `PseudoPotentialFamily` subclasses, but by necessity there will still be a large chunk of logic to deal with the detailed inconsistency in PseudoDojo.

sphuber:
> Why not support PK and UUID in `load_pseudo_family`? Supporting this here does not preclude loading the `Group` directly but adds the benefit of `load_pseudo_family` being a one-stop-shop for users, no matter what identifier they have on hand.

It is a good point. I mostly didn't do it because it complicates the implementation and behavior, but maybe you are right that having to switch the function based on what identifier you have is going to be annoying.

Thinking about this some more, I am not even sure anymore why we concluded that we needed the `pseudo_type` argument to have people distinguish pseudo families when they have a label. The uniqueness requirement in the database is on the tuple `(label, type_string)` which means that you cannot have two families of the same type with the same label. That means that if you want to create `PseudoPotentialFamily` instances with `UpfData` nodes, then you _have_ to use different labels anyway. So essentially the ambiguity problem that I thought was there, doesn't even exist, does it?

So that means we can just remove `pseudo_type` from the function, meaning we can simply get rid of it entirely, because at that point it is identical to `load_group` essentially... Am I missing something?

zooks97 (approved):

PR #33, opened 2020-12-07
`PseudoPotentialData`: add the `get_or_create` classmethod 

### LINKED ISSUE #32: Add `PseudoPotentialData.get_or_create` classmethod
Certain users of the old `UpfData` find the `get_or_create` classmethod useful to prevent creating a new node for a pseudopotential that is already contained within the database based on the md5 checksum. The `PseudoPotentialFamily.create_from_folder` already contains this deduplicating functionality, but if one wants to construct a single pseudopotential data node, this is not available.

sphuber:
Fixes #32 

The classmethod has the exact same signature as the constructor but
instead of constructing a new instance straightaway, it will first check
if the database already contains a node with the exact same class (so no
subclassing is used in the query) and with an md5 checksum that matches
that of the bytes of the stream.

If an existing node is matched, that is returned, otherwise a new
instance is constructed and returned unstored.

### DIFF
--- a/aiida_pseudo/data/pseudo/pseudo.py
+++ b/aiida_pseudo/data/pseudo/pseudo.py
@@ -2,21 +2,41 @@
 """Base class for data types representing pseudo potentials."""
 import typing
 
+from aiida import orm
+from aiida import plugins
 from aiida.common.constants import elements
 from aiida.common.exceptions import StoringNotAllowed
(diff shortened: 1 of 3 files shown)


zooks97 (approved):

bosonie (approved):
All good

PR #35, opened 2020-12-08
`RecommendedCutoffMixin`: require electronvolts for cutoffs

### LINKED ISSUE #15: Consistent units for suggested cutoffs
Currently, the SSSP is using Ry by default as the units of its suggested cutoffs (these are the units in the metadata JSON). This is relatively ok because UPF pseudopotentials are generally used by Q-E, which uses Ry for its energy cutoff input.

PseudoDojo provides suggested cutoffs in Ha, but it also provides its pseudopotentials in various formats (UPF for Q-E, PSML for Siesta/Abinit, and PSP8 for Abinit). 

Because a variety of codes with different unit conventions are supported, I feel that we should move to a standard unit for suggested cutoffs, probably eV, so that unit problems do not crop up in the future.

sphuber:
Fixes #15 

Since multiple codes might start to use the `RecommendedCutoffMixin`
which might each require the cutoffs in different units, it is important
to have a standard unit. For now, electronvolt is chosen as the one and
only supported unit. Reason being that electronvolt is the default
energy unit for most of the AiiDA ecosphere up till now. In the future,
the mixin may be updated to support multiple units with automatic
conversion but this can be added later in a backwards compatible manner.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,10 +23,12 @@ Installed pseudopotential families can be listed using:
 Any pseudopotential family installed can be loaded like any other `Group` using the `load_group` utility from `aiida-core`.
 Once loaded, it is easy to get the pseudopotentials for a given element or set of elements:
 
-    from aiida.orm import load_group
-    family = load_group('SSSP/1.1/PBE/efficiency')
-    pseudo = family.get_pseudo(element='Ga')  # Returns a single pseudo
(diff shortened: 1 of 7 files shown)


zooks97 (approved):
Looks good

PR #39, opened 2020-12-16
Remove support for PseudoDojo v0.3; enable fixed v0.4 families

zooks97:
The PseudoDojo team recommended to remove support for v0.3 because it is based on a bugged pseudopotential code, so I have removed it.

They have also fixed the remaining inconsistencies (F pseudopotential) in the v0.4 families. So, these families have been added to the valid configurations.


sphuber (approved):
All good, thanks @zooks97

Issue #40, opened 2020-12-16
Find method to fill in missing PseudoDojo PAW cutoffs or drop support? 


zooks97:
Unfortunately, the authors of the PseudoDojo PAW tables have said that official hints cannot be provided for the lanthanides due to unavailability of delta-factors calculated for these potentials.

The PseudoDojo team mentioned that 

> as a rule of thumb, the PAW lanthanides should require about 20 Ha, but consider this as a very qualitative estimate. Moreover, some of the PAW pseudos have 4f electrons in the valence, and this implies that:
> 
> 1. more advanced formalisms (e.g. LDA+U or DMFT) may be needed to capture the correct physics
> 2. One may need to increase the cutoff for the double grid (pawecutdg in Abinit) to correctly describe the on-site terms in real-space

It seems that for the pseudos that are missing cutoffs, it would be difficult to have even a heuristic estimate.

However, the elements which are missing cutoffs are (in general) also not present at all in the other PseudoDojo families. It could be possible to provide a (nearly) complete family if we drop the troublesome cases (with a warning, of course).

zooks97:
This seems to have become a more pressing issue as we need the PAW potentials in `aiida-common-workflows` to get EOS results which more closely match those of other codes.

sphuber:
since `abinit` can read UPF, why not use the `SSSP`? Then we are sure we are using the same pseudos? That is of course a short term solution but one that might work while you figure out the solution for recommended cutoffs for the PAW pseudos.

zooks97:
I don't think `abinit` can read `UPF`, although I could try it.

zooks97:
The maximum cutoff values for all PAW families are
```
low: 20 [Ha]
normal: 25 [Ha]
high: 25 [Ha]
```
I suggest using these values to fill in the missing cutoffs.
As recommended by the Abinit devs, I would the following for the PAW duals (`pawecutdg`)
```
low: 1.5
normal: 2.0
high: 2.0
```
@sponce24 , what do you think?

zooks97:
![image](https://user-images.githubusercontent.com/8887252/102514916-612f7a80-408d-11eb-9c4a-e2be8552b8dc.png)
This is the completeness of the PAW cutoffs. For reference, the norm-conserving v0.4 completeness is also shown.

sphuber:
Fixed in PR #42

PR #41, opened 2020-12-17
Add support for PseudoDojo PAW pseudos

zooks97:
Discussed in issue #40 

Because only the Lanthanides are missing cutoffs in the PseudoDojo PAW families, which in general are not available in the other PseudoDojo families, I simply remove the elements with missing (placeholder, value `-1`) cutoffs from the `PseudoDojoFamily` during installation and warn the user that I'm doing it.


sphuber:
Superseded by #42

PR #42, opened 2020-12-17
Add support for PseudoDojo PAW pseudos (alternative)

zooks97:
Discussed in issue #40 
Alternative to PR #41 

The range of recommended cutoffs in the PseudoDojo PAW families is relatively small (min 10, 10, 10; max 20, 20, 25 for low, normal, high stringency). So, it seems fairly reasonable to replace placeholder cutoff values (i.e. -1) with the maximum for the corresponding stringency. This is an alternative solution to dropping the pseudos entirely, as proposed in #41

### DIFF
--- a/aiida_pseudo/cli/install.py
+++ b/aiida_pseudo/cli/install.py
@@ -241,8 +241,8 @@ def cmd_install_pseudo_dojo(version, functional, relativistic, protocol, pseudo_
                 echo.echo_critical(msg)
 
         # The PAW configurations have missing cutoffs for the Lanthanides, which have ben replaced with a placeholder
-        # value of `-1`. We replace these with the maximum cutoff from the same stringency so that these potentials are
-        # still useable, but this should be taken as a _rough_ approximation.
+        # value of `-1`. We replace these with the 1.5 * the maximum cutoff from the same stringency so that these
+        # potentials are still usable, but this should be taken as a _rough_ approximation.
         # We check only the `cutoff_wfc` because `cutoff_rho` is not provided by PseudoDojo and is therefore
         # locally calculated in `PseudoDojoFamily.parse_djrepos_from_archive` as `2.0 * cutoff_wfc` for PAW.
         if configuration in paw_configurations:
@@ -250,16 +250,17 @@ def cmd_install_pseudo_dojo(version, functional, relativistic, protocol, pseudo_
             for stringency, str_cutoffs in cutoffs.items():
                 adjusted_cutoffs[stringency] = []
                 max_cutoff_wfc = max([str_cutoffs[element]['cutoff_wfc'] for element in str_cutoffs])
+                filler_cutoff_wfc = max_cutoff_wfc * 1.5
                 for element, cutoff in str_cutoffs.items():
                     if cutoff['cutoff_wfc'] <= 0:
-                        cutoffs[stringency][element]['cutoff_wfc'] = max_cutoff_wfc
-                        cutoffs[stringency][element]['cutoff_rho'] = 2.0 * max_cutoff_wfc
+                        cutoffs[stringency][element]['cutoff_wfc'] = filler_cutoff_wfc
+                        cutoffs[stringency][element]['cutoff_rho'] = 2.0 * filler_cutoff_wfc
                         adjusted_cutoffs[stringency].append(element)
 
             for stringency, elements in adjusted_cutoffs.items():
                 msg = f'stringency `{stringency}` has missing recommended cutoffs for elements ' \
-                    f'{", ".join(elements)}: as a substitute, the maximum cutoff of the stringency was set for these ' \
-                    'elements. USE WITH CAUTION!'
+                    f'{", ".join(elements)}: as a substitute, 1.5 * the maximum cutoff of the stringency ' \
+                    'was set for these elements. USE WITH CAUTION!'
                 echo.echo_warning(msg)
 
         family.description = description
--- a/aiida_pseudo/cli/install.py
+++ b/aiida_pseudo/cli/install.py
@@ -237,15 +237,15 @@ def cmd_install_pseudo_dojo(version, functional, relativistic, protocol, pseudo_
             max_cutoff_rho = max([str_cutoffs[element]['cutoff_rho'] for element in str_cutoffs])
             for element, cutoff in str_cutoffs.items():
                 if cutoff['cutoff_wfc'] <= 0:
-                    cutoffs[stringency][element]['cutoff_wfc'] = max_cutoff_wfc
+                    cutoffs[stringency][element]['cutoff_wfc'] = max_cutoff_wfc * 2.0
                     adjusted_cutoff_elements.append(element)
                 if cutoff['cutoff_rho'] <= 0:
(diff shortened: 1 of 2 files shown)


zooks97:
An even safer alternative suggested by @sponce24 is to set the unknown cutoffs to 1.5 ~ 2.0 times the maximum in the same stringency.

aiida_pseudo/cli/install.py
```diff
@@ -231,6 +231,23 @@ def cmd_install_pseudo_dojo(version, functional, relativistic, protocol, pseudo_
                 msg = f'md5 of pseudo for element {element} does not match that of the metadata {md5}'
                 echo.echo_critical(msg)
 
+        adjusted_cutoff_elements = []
```
sphuber, line 234: I wonder if we should explicit only do this logic for the PAW configurations? Such that we cannot accidentally start to apply this to other configurations that shouldn't have missing cutoffs? In any case, we should add a comment why this logic is here and what it is doing.

aiida_pseudo/cli/install.py
```diff
@@ -231,6 +231,23 @@ def cmd_install_pseudo_dojo(version, functional, relativistic, protocol, pseudo_
...
+        for stringency, str_cutoffs in cutoffs.items():
+            max_cutoff_wfc = max([str_cutoffs[element]['cutoff_wfc'] for element in str_cutoffs])
+            max_cutoff_rho = max([str_cutoffs[element]['cutoff_rho'] for element in str_cutoffs])
+            for element, cutoff in str_cutoffs.items():
+                if cutoff['cutoff_wfc'] <= 0:
+                    cutoffs[stringency][element]['cutoff_wfc'] = max_cutoff_wfc * 2.0
+                    adjusted_cutoff_elements.append(element)
+                if cutoff['cutoff_rho'] <= 0:
```
sphuber, line 242: Wonder if we should add a check that if we are adjusting the wfc cutoff, we also have to update the rho cutoff. At least that way we make sure that they stay consistent.

aiida_pseudo/cli/install.py
sphuber, line 234: probably make this a dictionary with list for each stringency, such that in the warnings you can say explicitly for which stringency we had to make adjustments

aiida_pseudo/cli/install.py
```diff
@@ -231,6 +231,23 @@ def cmd_install_pseudo_dojo(version, functional, relativistic, protocol, pseudo_
                 msg = f'md5 of pseudo for element {element} does not match that of the metadata {md5}'
                 echo.echo_critical(msg)
 
+        adjusted_cutoff_elements = []
+        for stringency, str_cutoffs in cutoffs.items():
+            max_cutoff_wfc = max([str_cutoffs[element]['cutoff_wfc'] for element in str_cutoffs])
+            max_cutoff_rho = max([str_cutoffs[element]['cutoff_rho'] for element in str_cutoffs])
+            for element, cutoff in str_cutoffs.items():
+                if cutoff['cutoff_wfc'] <= 0:
+                    cutoffs[stringency][element]['cutoff_wfc'] = max_cutoff_wfc * 2.0
```
sphuber, line 240: If we are already taking the maximum of the stringency set, should we really even double it? Can't find your comment anymore about what the PD team recommended about cutoffs for lanthanides. I have the feeling this is a bit excessive. If we give a warning that they are not officially tested, maybe just using the max itself should be fine?

aiida_pseudo/cli/install.py
```diff
@@ -231,6 +231,23 @@ def cmd_install_pseudo_dojo(version, functional, relativistic, protocol, pseudo_
...
+                    adjusted_cutoff_elements.append(element)
+                if cutoff['cutoff_rho'] <= 0:
+                    cutoffs[stringency][element]['cutoff_rho'] = max_cutoff_rho * 2.0
+                    adjusted_cutoff_elements.append(element)
+
+        for element in set(adjusted_cutoff_elements):
+            msg = f'element {element} had invalid cutoff(s) which were modified to two times the maximum cutoff ' \
+                'value of the same stringency'
+            echo.echo_warning(msg)
```
sphuber, line 249:
```suggestion
        for stringency, elements in adjusted_cutoff_elements.items():
            msg = f'stringency `{stringency}` has missing recommended cutoffs for elements {', '.join(elements)}: as a substitute, the maximum cutoff of the stringency was set for these elements.'
            echo.echo_warning(msg)
```

sphuber (changes requested):
Thanks @zooks97 . I am happy with the concept, just some suggestions on implementation and comments/messaging

aiida_pseudo/cli/install.py
zooks97: I can do this only for PAW configurations and add a comment

zooks97: All elements that are missing cutoffs are missing them for all stringencies. So, this will make the output three times as verbose, but we can do it if you like.

aiida_pseudo/cli/install.py
zooks97: They recommended 20 Ha as a rule of thumb, but they also mentioned that the density (PAW double grid) cutoff may need to be larger than what is generally required. As the elements that are missing are only the Lanthanides, it could make sense to do something "too strict" to be safe.

aiida_pseudo/cli/install.py
sphuber: I realize that, but that is why I reformatted the warning messages to have a single message per stringency. This way there will be three lines with a warning at most, which I think is acceptable. This way we also are clear in the case in the future somehow the cutoffs are only missing for certain stringency sets

aiida_pseudo/cli/install.py
sphuber: ok, but what is the max cutoff in the strict set? Is that more or less than 20 Ha. Because it if iis more, and we times two that, it will be _a lot_ more than 20 Ha, so maybe a bit excessive.

aiida_pseudo/cli/install.py
zooks97: Because I generate the `cutoff_rho` as `2.0 * cutoff_wfc` for the PAW potentials, I really only need to check if the `cutoff_wfc <= 0`

zooks97:
Note: I've changed the `dual` for the norm-conserving families to 4.0 after a conversation with @sponce24 . The previous value of 8.0 was unnecessary and wasteful.

aiida_pseudo/cli/install.py
zooks97: I think the max in the strict set is 25, so I agree that 2.0 would be a bit extreme. Perhaps a lower multiplier would be better.

zooks97:
@sphuber I can't remember which solution we decided upon in the end, and Marnik has run into an issue with the incorrect dual value which was fixed in this pull. Could we think about merging one of the solutions?

sphuber:
We agreed on this variant which indeed already fixes the dual factor. The only thing remaining was the behavior for setting recommended cutoffs for those that don't have one. I thought that twice the max may be a bit much (see the only remaining open comment). Should we set the factor to 1.5 ? If you agree and update, then I will merge

sphuber (approved):

PR #44, opened 2021-02-10
Add support for OpenMX VPS format

zooks97:
I'm working on implementing an AiiDA plugin for the O(N)/local numerical pseudo atomic orbital basis DFT code [OpenMX](http://www.openmx-square.org) over in [`aiida-openmx`](https://github.com/zooks97/aiida-openmx), so I've added support for their rather eclectic VPS pseudopotential format.

The element / valence / exchange-correlation parsing is rather brittle as a general rule because it's all based on the copied input file that they include as a header to all their VPS files. I'm not sure if their pseudopotential code puts this in by itself or if that's just how they prepare the files for release.

As far as I can tell, there is no information about the atomic species in the actual pseudopotential block at the bottom of the VPS files, so this is the best way I've found to get it.

I think that personally-generated pseudopotentials for this code are so niche that any problems that arise can be dealt with when / if they do.

I didn't add any families / CLI installer for these because the pseudos are only available bundled with the source, and I haven't taken the time to codify the suggestions [table](http://www.openmx-square.org/openmx_man3.9/node27.html) that they provide.


zooks97:
I didn't change the version in the commit, but it would be nice to have at least a minor version bump with this.

zooks97:
Pinging @sphuber : I'm moving towards an `aiida-openmx` release in the next few days, and I depend on `VpsData` for the pseudopotentials there.

sphuber:
will try to look at it by tomorrow evening

sphuber:
@zooks97 thanks, this all looks fine. If you fix the pre-commit I will merge this.

zooks97:
I didn't change anything in the file where the pre-commit failed. It says something about using a generator in conftest.py

sphuber:
It's because a new version of `pylint` apparently got released that adds a new check. All you have to do is remove the square brackets from line 145 in `tests/conftest.py` and it should work.

zooks97:
All good!

sphuber (approved):
Thanks @zooks97

PR #52, opened 2021-03-17
CLI: add recommended cutoffs to `aiida-pseudo family show`

### LINKED ISSUE #51: Display recommended cutoffs in `aiida-pseudo family show` if present

sphuber:
Fixes #51 

Previously, `aiida-pseudo family show` merely shows the element,
filename and md5 of each pseudo contained within a family. Since the
addition of the `RecommendedCutoffMixin`, families can also specify
recommended cutoffs. Here we add these to the output of the command if
the specified family is a subclass of `RecommendedCutoffMixin`.

In order to test this, a generic pseudo potential family class with the
`RecommendedCutoffMixin` is needed. This was already implemented for the
tests of the mixin itself, which created a new group class on the fly
using the `aiida.orm.groups.GroupMeta` meta class to automatically set
the required type strings. This worked for these tests since they didn't
have to be queried for, however, for the `aiida-pseudo family show`
tests, the CLI command will have to query for it and since the mock
class does not have an actual physical entry point, the query fails.

Since dynamically adding an entry point in the setup of the test suite
is not trivial, instead we add an actual `CutoffsFamily` class including
an entry point. This class is designed therefore to be mostly used for
testing purposes.

### DIFF
--- a/aiida_pseudo/cli/show.py
+++ b/aiida_pseudo/cli/show.py
@@ -5,22 +5,35 @@ import click
 from aiida.cmdline.params import options as options_core
 from aiida.cmdline.utils import decorators, echo
 
-from .params import PseudoPotentialFamilyParam
+from ..groups.mixins import RecommendedCutoffMixin
+from .params import PseudoPotentialFamilyParam, options
 from .root import cmd_root
 
 
 @cmd_root.command('show')
 @click.argument('family', type=PseudoPotentialFamilyParam(sub_classes=('aiida.groups:pseudo.family',)))
+@options.STRINGENCY()
 @options_core.RAW()
 @decorators.with_dbenv()
-def cmd_show(family, raw):
+def cmd_show(family, stringency, raw):
     """Show details of pseudo potential family."""
     from tabulate import tabulate
 
-    rows = [[pseudo.element, pseudo.filename, pseudo.md5] for pseudo in family.nodes]
-    headers = ['Element', 'Pseudo', 'MD5']
+    if isinstance(family, RecommendedCutoffMixin):
+
+        if stringency is not None and stringency not in family.get_cutoff_stringencies():
+            raise click.BadParameter(f'`{stringency}` is not defined for family `{family}`.', param_hint='stringency')
+
+        headers = ['Element', 'Pseudo', 'MD5', 'Wavefunction (eV)', 'Charge density (eV)']
+        rows = [[
+            pseudo.element, pseudo.filename, pseudo.md5,
+            *family.get_recommended_cutoffs(elements=pseudo.element, stringency=stringency)
+        ] for pseudo in family.nodes]
+    else:
+        headers = ['Element', 'Pseudo', 'MD5']
+        rows = [[pseudo.element, pseudo.filename, pseudo.md5] for pseudo in family.nodes]
 
     if raw:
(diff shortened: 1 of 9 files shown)


aiida_pseudo/cli/show.py
```diff
@@ -17,10 +18,15 @@ def cmd_show(family, raw):
     """Show details of pseudo potential family."""
     from tabulate import tabulate
 
-    rows = [[pseudo.element, pseudo.filename, pseudo.md5] for pseudo in family.nodes]
-    headers = ['Element', 'Pseudo', 'MD5']
+    if isinstance(family, RecommendedCutoffMixin):
+        headers = ['Element', 'Pseudo', 'MD5', 'Wavefunction', 'Charge density']
```
mbercx, line 22: Should we also add the units to the cutoff headers here? For now it'd just be `eV`, but once we add support for other units these can be extracted from the family (e.g. using your suggested `get_recommended_cutoff_unit` method).

mbercx (approved):
Thanks @sphuber! Changes look good, I just have two suggestions:

* Adding units to the headers of the table. (see comment)
* Adding `stringency` option to the `aiida-pseudo family show` which is passed to the `get_recommended_cutoffs` method.

Since both are minor and non-essential, I'll already approve.

aiida_pseudo/cli/show.py
sphuber: Yes, good point. I was thinking of adding this when we address the issue of storing the unit explicitly in the extras, but might already hardcode it here that it is eV. I was already working on the stringency and it is almost done. Will push changes soon

sphuber:
On another note, I am working on another PR that allows setting the cutoffs manually from the CLI. I feel a `aiida-pseudo family` command group would make most sense for this, but to my surprise (and yours apparently as well, since you used `aiida-pseudo family show` in your comment) does not yet exist. It is `aiida-pseudo show` to show details of a family. If you agree as well I would move this to `aiida-pseudo family show` and then I can add there `aiida-pseudo family set-cutoffs` or something like that

mbercx:
> but to my surprise (and yours apparently as well, since you used `aiida-pseudo family show` in your comment) does not yet exist.

Hehe, I must admit I simply copied the command from the PR title. 😅  However, I do agree that `aiida-pseudo family show` would be more intuitive, especially when adding more commands like `set-cutoffs`.

sphuber:
I will make that change in a separate PR though. Would be good to merge PR #55 first as it already puts `aiida-pseudo family` in place. After that I can simply move the relevant code there

tests/conftest.py
```diff
@@ -123,7 +123,9 @@ def _get_pseudo_family(
         label='family',
         cls=PseudoPotentialFamily,
         pseudo_type=PseudoPotentialData,
-        elements=None
+        elements=None,
+        cutoffs=None,
```
mbercx, line 127: Maybe still add `cutoffs` and `default_stringency` to the docstring? These input arguments also don't seem to be used anywhere at the moment (unless I missed it). 😅

mbercx (approved):
All good! Just have one minor comment re the `get_pseudo_family` fixture. Something I also noticed is that I can't tab-complete the pseudo family, which might be a nice addition now that we have commands where you have to provide the (often long) label as an argument. But maybe you've already added that in the other PR's I still have to review. 👀

PR #54, opened 2021-03-17
Dependencies: temporarily limit upper version of `sqlalchemy`

### LINKED ISSUE #53: Temporarily pin `sqlalchemy` due to incompatibility of `v1.4` with `sqlalchemy-utils`
See this issue https://github.com/kvesteri/sqlalchemy-utils/issues/505

sphuber:
Fixes #53 

The newly released `sqlalchemy==1.4.0` breaks `sqlalchemy-utils` and so
breaks all the tests. Until the latter releases a fix, we have to put an
upper limit on the `sqlalchemy` dependency.

### DIFF
--- a/setup.json
+++ b/setup.json
@@ -38,7 +38,8 @@
         "aiida-core~=1.4",
         "click~=7.0",
         "click-completion~=0.5",
-        "requests~=2.20"
+        "requests~=2.20",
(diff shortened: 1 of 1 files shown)


mbercx (approved):
Thanks @sphuber!

PR #55, opened 2021-03-17
CLI: add the `aiida-pseudo family cutoffs set` command

### LINKED ISSUE #50: Allow setting of recommended cutoff parameters through the CLI
The CLI provides `aiida-pseudo install family` to manually install a family from an archive on disk. This is a backup solution in the case the automated solutions for the PD and SSSP don't work. However, this method does not allow to set the recommended cutoffs, making the family useless for the automated protocols since they cannot automatically determine the cutoffs for a given structure.

sphuber:
Fixes #50 

This command allows to set recommended cutoffs for a given family. The
cutoffs are specified in a JSON file that is given as an argument and
the stringency is specified through the `-s/--stringency` option. If the
stringency already exists, it is simply overwritten.

### DIFF
--- a/aiida_pseudo/cli/family.py
+++ b/aiida_pseudo/cli/family.py
@@ -1,0 +1,58 @@
+# -*- coding: utf-8 -*-
+"""Commands to inspect or modify the contents of pseudo potential families."""
+import json
+
+import click
+
+from aiida.cmdline.utils import decorators, echo
+
+from ..groups.mixins import RecommendedCutoffMixin
+from .params import arguments, options
+from .root import cmd_root
+
+
+@cmd_root.group('family')
+def cmd_family():
+    """Command group to inspect or modify the contents of pseudo potential families."""
+
+
+@cmd_family.group('cutoffs')
+def cmd_family_cutoffs():
+    """Command group to inspect or modify the recommended cutoffs of pseudo potential families."""
+
+
+@cmd_family_cutoffs.command('set')
+@arguments.PSEUDO_POTENTIAL_FAMILY()
+@click.argument('cutoffs', type=click.File(mode='rb'))
+@options.STRINGENCY(required=True)
+@decorators.with_dbenv()
+def cmd_family_cutoffs_set(family, cutoffs, stringency):  # noqa: D301
+    """Set the recommended cutoffs for a pseudo potential family.
+
+    The cutoffs should be provided as a JSON file through the argument `CUTOFFS` which should have the structure:
+
+    \b
+        {
+            "Ag": {
+                "cutoff_wfc": 50.0,
+                "cutoff_rho": 200.0
+            },
+            ...
+        }
+
+    where the cutoffs are expected to be in electronvolt.
+    """
+    if not isinstance(family, RecommendedCutoffMixin):
+        raise click.BadParameter(f'family `{family}` does not support recommended cutoffs to be set.')
+
+    try:
+        data = json.load(cutoffs)
+    except ValueError as exception:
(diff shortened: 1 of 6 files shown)


sphuber:
Was wondering if we might need to emit a warning if a stringency is specified that already exists, with the `--force` option to prevent prompting. A lot more work and not sure if it is really necessary.

sphuber:
At some point will probably also need a `family cutoffs default` and `family cutoffs list`

mbercx:
> Was wondering if we might need to emit a warning if a stringency is specified that already exists, with the `--force` option to prevent prompting. A lot more work and not sure if it is really necessary.

I'd think someone that goes through the effort of preparing the `json` file and then specifically writes an already existing stringency knows what they are doing?

aiida_pseudo/cli/family.py
```diff
@@ -0,0 +1,58 @@
...
+    """Command group to inspect or modify the recommended cutoffs of pseudo potential families."""
+
+
+@cmd_family_cutoffs.command('set')
+@arguments.PSEUDO_POTENTIAL_FAMILY()
+@click.argument('cutoffs', type=click.File(mode='rb'))
+@options.STRINGENCY(required=True)
+@decorators.with_dbenv()
+def cmd_family_cutoffs_set(family, cutoffs, stringency):
+    r"""Set the recommended cutoffs for a pseudo potential family.
```
mbercx, line 30: The raw format messes up your beautiful formatting 😭 

```suggestion
    """Set the recommended cutoffs for a pseudo potential family.
```

aiida_pseudo/cli/family.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+@cmd_family.group('cutoffs')
+def cmd_family_cutoffs():
+    """Command group to inspect or modify the recommended cutoffs of pseudo potential families."""
+
+
+@cmd_family_cutoffs.command('set')
+@arguments.PSEUDO_POTENTIAL_FAMILY()
+@click.argument('cutoffs', type=click.File(mode='rb'))
```
mbercx, line 26: Is there a reason for making the cutoffs an argument and the stringency a required option?

mbercx (approved):
Thanks @sphuber! Just added a small fix for the command docstring and a question.

One use case I'm still considering is when someone wants to change the cutoffs only for a specific element. This is currently not possible because all the elements need to provided to the `set_cutoffs` method. However, I can imagine I'm working on a specific class of materials and just want to increase the cutoff for oxygen for example. Would it make sense to allow this in case the stringency specified is already defined, and then just load the cutoffs and override those provided in the JSON by the user?

aiida_pseudo/cli/family.py
sphuber: Hmm, in the current conception. I think originally I intended to allow to specify multiple stringencies at once and then let one define the default. But indeed, now that only on stringency is defined at a time, maybe it makes more sense to specify it as an argument. On the other hand, leaving it like this does make it possible in the future to make the command more flexible and also allow a file with multiple stringencies defined, where the keys are the stringency name. Then it makes more sense to make the default stringency parameter an option and use the first one defined in the file as the default such that the flag becomes optional. Think that is the best option actually thinking about it now

aiida_pseudo/cli/family.py
sphuber: Weird, I could have sworn that I added this because without it the `\b` marker didn't seem to be respected. But indeed, removing it keeps the formatting as intended

sphuber: Ah I see why, it is `pydocstyle` that complains about the `\b` marker saying I should use `r` when using backslashes. Will try to add an exception

PR #57, opened 2021-03-17
`RecommendedCutoffMixin`: store the unit of cutoffs in extras

### LINKED ISSUE #47: Storing the units for the recommended cuttoffs
Recently when I was trying to use aiida-pseudo with https://github.com/aiidateam/aiida-quantumespresso-epfl, it turned out I needed to convert the units from those I'd previously used with aiida-sssp, and also @flavianojs got into some similar issues with units.

So it would be good if the units were stored as part of this data, and in some way e.g. aiida-quantumespresso could check this (converting if necessary or raising an error if it can't)
(alternatively it could directly store the conversion factor to hartree or eV)

cc also @mbercx

sphuber:
Fixes #47 

By storing the unit of the cutoffs values explicitly in the extras, it
not only will give more certainty to the user of what the cutoffs mean
because they can inspect the unit, but the `get_recommended_cutoffs`
method can also provide the `unit` argument to automatically convert it
to another unit of energy if so desired.

The conversion between units is provided by the `pint` library, which is
added as a dependency. Users can specify units in the API through
strings using the names that are recognized by `pint`. Both the
shorthand, e.g. `eV`, as well as the long version `electron_volt` can be
used. Other common energy units are `Ry` and `Eh`, for Rydberg and
Hartree atomic units respectively.

The default unit that is set when defining recommended cutoffs is still
set to be electron volt, which was the implicitly assumed unit uptil now.
Families that have been created with versions preceding this commit,
will not actually have the unit defined in the unit, but
`get_cutoffs_unit` will return the default electron volt if it is
missing. This does mean that the `RecommendedCutoffMixin.DEFAULT_UNIT`
cannot be changed or this would lead to old families to be interpreted
to have the incorrect unit.

### DIFF
--- a/aiida_pseudo/cli/install.py
+++ b/aiida_pseudo/cli/install.py
@@ -84,7 +84,6 @@ def cmd_install_sssp(version, functional, protocol, traceback):
     from aiida.orm import Group, QueryBuilder
 
     from aiida_pseudo import __version__
-    from aiida_pseudo.common import units
     from aiida_pseudo.groups.family import SsspConfiguration, SsspFamily
     from .utils import attempt, create_family_from_archive
 
@@ -136,13 +135,10 @@ def cmd_install_sssp(version, functional, protocol, traceback):
                 echo.echo_critical(msg)
 
             # Cutoffs are in Rydberg but need to be stored in the family in electronvolt.
-            cutoffs[element] = {
-                'cutoff_wfc': values['cutoff_wfc'] * units.RY_TO_EV,
-                'cutoff_rho': values['cutoff_rho'] * units.RY_TO_EV,
-            }
+            cutoffs[element] = {'cutoff_wfc': values['cutoff_wfc'], 'cutoff_rho': values['cutoff_rho']}
 
         family.description = description
-        family.set_cutoffs({'normal': cutoffs})
+        family.set_cutoffs({'normal': cutoffs}, unit='Ry')
 
         echo.echo_success(f'installed `{label}` containing {family.count()} pseudo potentials')
 
@@ -264,6 +260,6 @@ def cmd_install_pseudo_dojo(version, functional, relativistic, protocol, pseudo_
                 echo.echo_warning(msg)
 
         family.description = description
-        family.set_cutoffs(cutoffs, default_stringency=default_stringency)
+        family.set_cutoffs(cutoffs, default_stringency=default_stringency, unit='Eh')
 
         echo.echo_success(f'installed `{label}` containing {family.count()} pseudo potentials')
--- a/aiida_pseudo/groups/mixins/cutoffs.py
+++ b/aiida_pseudo/groups/mixins/cutoffs.py
@@ -5,6 +5,8 @@ from typing import Union
 from aiida.common.lang import type_check
 from aiida.plugins import DataFactory
 
+from aiida_pseudo.common.units import U
+
 StructureData = DataFactory('structure')  # pylint: disable=invalid-name
 
 __all__ = ('RecommendedCutoffMixin',)
@@ -17,11 +19,14 @@ class RecommendedCutoffMixin:
     functions and the charge density. The units have to be in electronvolt.
     """
 
+    DEFAULT_UNIT = 'eV'
+
     _key_cutoffs = '_cutoffs'
+    _key_cutoffs_unit = '_cutoffs_unit'
     _key_default_stringency = '_default_stringency'
 
-    @classmethod
-    def validate_cutoffs(cls, elements: set, cutoffs: dict) -> None:
+    @staticmethod
+    def validate_cutoffs(elements: set, cutoffs: dict) -> None:
         """Validate a cutoff dictionary for a given set of elements.
 
         :param elements: set of elements for which to validate the cutoffs dictionary.
@@ -56,6 +61,22 @@ class RecommendedCutoffMixin:
                     raise ValueError(
                         f'invalid cutoff values for stringency `{stringency}` and element {element}: {values}'
                     )
+
+    @staticmethod
+    def validate_cutoffs_unit(unit: str) -> None:
+        """Validate the cutoffs unit.
+
+        The unit should be a name that is recognized by the ``pint`` library to be a unit of energy.
+
+        :raises ValueError: if an invalid unit is specified.
+        """
+        type_check(unit, str)
+
+        if unit not in U:
+            raise ValueError(f'`{unit}` is not a valid unit.')
+
+        if not U.Quantity(1, unit).check('[energy]'):
+            raise ValueError(f'`{unit}` is not a valid energy unit.')
 
     def validate_stringency(self, stringency: str) -> None:
         """Validate a cutoff stringency.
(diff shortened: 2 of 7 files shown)


sphuber:
@mbercx I served you a cold one! :beer:

mbercx (changes requested):
Deze PR is bijna even lekker als een Tripel Karmeliet! 🍺 

Still left a few comments. Main issue is the fact that the units are set to `None` in the extras when no unit is specified in the `set_cutoffs` method.

aiida_pseudo/groups/mixins/cutoffs.py
```diff
@@ -107,16 +128,19 @@ def set_cutoffs(self, cutoffs: dict, default_stringency: str = None) -> None:
         :param default_stringency: the default stringency to be used when ``get_recommended_cutoffs`` is called. If is
             possible to not specify this if and only if the cutoffs only contain a single stringency set. That one will
             then automatically be set as default.
+        :param unit: TODO
```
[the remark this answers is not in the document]
sphuber: external-content duckduckgo com

tests/groups/mixins/test_cutoffs.py
```diff
@@ -217,3 +230,45 @@ def test_get_recommended_cutoffs(get_pseudo_family, generate_structure):
...
+                'cutoff_wfc': 1.0,
+                'cutoff_rho': 2.0
+            },
+            'He': {
+                'cutoff_wfc': 3.0,
+                'cutoff_rho': 8.0
+            },
+        }
+    }
+    family = get_pseudo_family(label='SSSP/1.0/PBE/efficiency', cls=CutoffFamily, elements=elements)
```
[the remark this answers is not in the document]
sphuber: Hmm, just have to add the possibility to set the `unit` as well.

aiida_pseudo/cli/install.py
```diff
@@ -264,6 +260,6 @@ def cmd_install_pseudo_dojo(version, functional, relativistic, protocol, pseudo_
                 echo.echo_warning(msg)
 
         family.description = description
-        family.set_cutoffs(cutoffs, default_stringency=default_stringency)
+        family.set_cutoffs(cutoffs, default_stringency=default_stringency, unit='Eh')
```
mbercx, line 263: I suppose this is a typo?

```suggestion
        family.set_cutoffs(cutoffs, default_stringency=default_stringency, unit='Ev')
```

tests/groups/mixins/test_cutoffs.py
```diff
@@ -139,7 +164,7 @@ def test_set_cutoffs_auto_default(get_pseudo_family):
 
 @pytest.mark.usefixtures('clear_db')
 def test_get_cutoffs(get_pseudo_family):
-    """Test the `RecommendedCutoffMixin.get_cutoffs` method."""
+    """Test the `CutoffsFamily.get_cutoffs` method."""
```
mbercx, line 167: I couldn't resist
```suggestion
    """Test the ``CutoffsFamily.get_cutoffs`` method."""
```

tests/groups/mixins/test_cutoffs.py
```diff
@@ -157,7 +182,7 @@ def test_get_cutoffs(get_pseudo_family):
 
 @pytest.mark.usefixtures('clear_db')
 def test_get_recommended_cutoffs(get_pseudo_family, generate_structure):
-    """Test the `RecommendedCutoffMixin.get_recommended_cutoffs` method."""
+    """Test the `CutoffsFamily.get_recommended_cutoffs` method."""
```
mbercx, line 185:
```suggestion
    """Test the ``CutoffsFamily.get_recommended_cutoffs`` method."""
```

tests/groups/mixins/test_cutoffs.py
```diff
@@ -201,3 +231,51 @@ def test_get_recommended_cutoffs(get_pseudo_family, generate_structure):
...
+    expected = (cutoffs_ar['cutoff_wfc'], cutoffs_ar['cutoff_rho'])
+    assert family.get_recommended_cutoffs(elements='Ar') == expected
+
+    expected = (cutoffs_ar['cutoff_wfc'] * 2, cutoffs_ar['cutoff_rho'] * 2)
+    assert family.get_recommended_cutoffs(elements='Ar', unit='Ry') == expected
+
+
+@pytest.mark.usefixtures('clear_db')
+def test_get_cutoffs_unit(get_pseudo_family, get_cutoffs):
+    """Test the `CutoffsFamily.get_cutoffs_unit` method."""
```
mbercx, line 273:
```suggestion
    """Test the ``CutoffsFamily.get_cutoffs_unit`` method."""
```

tests/groups/mixins/test_cutoffs.py
```diff
@@ -84,7 +97,7 @@ def test_get_cutoff_stringencies(get_pseudo_family, get_cutoffs):
 
 @pytest.mark.usefixtures('clear_db')
 def test_set_cutoffs(get_pseudo_family):
-    """Test the `RecommendedCutoffMixin.set_cutoffs` method."""
+    """Test the `CutoffsFamily.set_cutoffs` method."""
```
mbercx, line 100: The OCD! It burns!
```suggestion
    """Test the ``CutoffsFamily.set_cutoffs`` method."""
```

tests/groups/mixins/test_cutoffs.py
```diff
@@ -28,17 +28,30 @@ def _get_cutoffs(family, stringencies=('default',)):
...
     family = get_pseudo_family(cls=CutoffsFamily)
     assert family._get_cutoffs() == {}  # pylint: disable=protected-access
 
     family.set_cutoffs(get_cutoffs(family), 'default')
     assert family._get_cutoffs() == get_cutoffs(family)  # pylint: disable=protected-access
 
 
+@pytest.mark.usefixtures('clear_db')
+def test_validate_cutoffs_unit():
+    """Test the `CutoffsFamily.validate_cutoffs_unit` method."""
```
mbercx, line 41:
```suggestion
    """Test the ``CutoffsFamily.validate_cutoffs_unit`` method."""
```

tests/conftest.py
```diff
@@ -125,13 +125,15 @@ def _get_pseudo_family(
         pseudo_type=PseudoPotentialData,
         elements=None,
         cutoffs=None,
+        unit=None,
         default_stringency=None
     ) -> PseudoPotentialFamily:
         """Return an instance of `PseudoPotentialFamily` or subclass containing the given elements.
```
mbercx, line 131: I have a problem
```suggestion
        """Return an instance of ``PseudoPotentialFamily`` or subclass containing the given elements.
```

aiida_pseudo/groups/mixins/cutoffs.py
```diff
@@ -143,13 +169,17 @@ def get_recommended_cutoffs(self, *, elements=None, structure=None, stringency=N
         :param stringency: optional stringency if different from the default.
         :return: tuple of recommended wavefunction and density cutoff.
```
mbercx, line 170: Where is the `unit` parameter? 😁

mbercx (approved):
Another round, bartender!

aiida_pseudo/cli/install.py
sphuber: Uhm, no? I think the cutoffs of the Pseudo Dojo tables are in Hartree, aren't they?

mbercx: Oh right 🤦  I forgot about `Eh` being an alternative for `Ha`. Seems `pint` also doesn't accept `Ha`.

mbercx (approved):
⭐ ⭐ ⭐ ⭐ ⭐  Would review again.

PR #65, opened 2021-04-09
CLI: add `--download-only` option to automatic family install commands

### LINKED ISSUE #63: CLI: add possibility to download pseudo archives and metadata files without installing
This would be useful in the case where the automated command fails and a user needs to perform a manual install through `aiida-pseudo install family`. Here it would be good to make downloading the files as easy as possible by either providing the URLs that need to be downloaded, or by downloading them through a command (but skipping the installation part).

sphuber:
Fixes #63 

The `aiida-pseudo install sssp` and `aiida-pseudo install pseudo-dojo`
automatically download the required files for installing the
corresponding pseudopotential families. Since the entire operation can
sometimes fail, `aiida-pseudo install family` was added in order to
facilitate the manual installation without the automatic download.
Conversely, it may be useful to have the command only download the files
of a particular family, without actually installing them.

The `--download-only` option is added to both automatic family install
commands which will simply write the files to the current working
directory after having downloaded them.

### DIFF
--- a/aiida_pseudo/cli/params/options.py
+++ b/aiida_pseudo/cli/params/options.py
@@ -9,7 +9,7 @@ from .types import PseudoPotentialFamilyTypeParam
 
 __all__ = (
     'VERSION', 'FUNCTIONAL', 'RELATIVISTIC', 'PROTOCOL', 'PSEUDO_FORMAT', 'STRINGENCY', 'DEFAULT_STRINGENCY',
-    'TRACEBACK', 'FAMILY_TYPE', 'ARCHIVE_FORMAT'
+    'TRACEBACK', 'FAMILY_TYPE', 'ARCHIVE_FORMAT', 'UNIT', 'DOWNLOAD_ONLY'
 )
 
 VERSION = OverridableOption(
@@ -81,4 +81,13 @@ UNIT = OverridableOption(
     default='eV',
     show_default=True,
     help='Specify the energy unit of the cutoffs. Must be recognized by the ``UnitRegistry`` of the ``pint`` library.'
+)
+
+DOWNLOAD_ONLY = OverridableOption(
+    '--download-only',
+    is_flag=True,
+    help=(
+        'Only download the pseudopotential files to the current working directory, without installing the '
+        'pseudopotential family.'
+    )
 )
--- a/aiida_pseudo/groups/family/pseudo_dojo.py
+++ b/aiida_pseudo/groups/family/pseudo_dojo.py
@@ -5,6 +5,8 @@ import json
 import os
 import re
 import warnings
+
+from pathlib import Path
 from typing import Sequence
 
 from aiida.common.exceptions import ParsingError
@@ -314,7 +316,7 @@ class PseudoDojoFamily(RecommendedCutoffMixin, PseudoPotentialFamily):
         return md5s, cutoffs
 
     @classmethod
-    def parse_djrepos_from_archive(cls, filepath_metadata, fmt=None, pseudo_type=None):
+    def parse_djrepos_from_archive(cls, filepath_metadata: Path, fmt=None, pseudo_type=None):
         """Parse metadata from a djrepo .tgz archive.
 
         .. warning:: the archive should not contain any subdirectories, but just the djrepo files.
--- a/tests/cli/test_install.py
+++ b/tests/cli/test_install.py
@@ -113,7 +113,7 @@ def run_monkeypatched_install_pseudo_dojo(run_cli_command, get_pseudo_potential_
 
 @pytest.mark.usefixtures('clear_db')
 def test_install_family(run_cli_command, get_pseudo_archive):
-    """Test `aiida-pseudo install family`."""
+    """Test ``aiida-pseudo install family``."""
     label = 'family'
     description = 'description'
     filepath_archive = next(get_pseudo_archive())
@@ -131,10 +131,10 @@ def test_install_family(run_cli_command, get_pseudo_archive):
 
 @pytest.mark.usefixtures('clear_db')
 def test_install_family_url(run_cli_command):
-    """Test `aiida-pseudo install family` when installing from a URL.
+    """Test ``aiida-pseudo install family`` when installing from a URL.
 
-    When a URL is passed, the parameter converts it into a `http.client.HTTPResponse`, which is not trivial to mock so
-    instead we use an actual URL, which is slow, but is anyway already tested indirectly in `test_install_sssp`.
+    When a URL is passed, the parameter converts it into a ``http.client.HTTPResponse``, which is not trivial to mock so
+    instead we use an actual URL, which is slow, but is anyway already tested indirectly in ``test_install_sssp``.
     """
     label = 'SSSP/1.0/PBE/efficiency'
     description = 'description'
@@ -153,7 +153,7 @@ def test_install_family_url(run_cli_command):
 
 @pytest.mark.usefixtures('clear_db')
 def test_install_sssp(run_cli_command):
-    """Test the `aiida-pseudo install sssp` command."""
+    """Test the ``aiida-pseudo install sssp`` command."""
     from aiida_pseudo import __version__
 
     result = run_cli_command(cmd_install_sssp)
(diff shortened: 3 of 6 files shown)


sphuber:
@mbercx have not yet added the tests but wanted to first get the design right. One major question in implementation is whether we should check whether the files already exist in the current working directory, or if we silently override as is currently happening, which is probably not a good idea. Just not sure what the behavior should be then. In case of clash simply print error and exit or add a force flag? It could be annoying if only one file already exists. If the second already existed, the first will be downloaded without problem before the command exits. If the user then deletes the second file from the warning and reruns the command, now it will fail again, because now the first file will already exist. I can see quite some annoying trial and error situations here.

mbercx:
> @mbercx have not yet added the tests but wanted to first get the design right. One major question in implementation is whether we should check whether the files already exist in the current working directory, or if we silently override as is currently happening, which is probably not a good idea. Just not sure what the behavior should be then. In case of clash simply print error and exit or add a force flag? It could be annoying if only one file already exists. If the second already existed, the first will be downloaded without problem before the command exits. If the user then deletes the second file from the warning and reruns the command, now it will fail again, because now the first file will already exist. I can see quite some annoying trial and error situations here.

Maybe we can just skip files that already exist and print a warning? I suppose these filenames are rather long and specific, so it's unlikely that a file with different content will be present in the current working directory. If the user does have a file with the same name in the directory by accident somehow, he/she will see the warning and rename/move the file and then simply execute the command again without issue.

mbercx:
Some quick field testing results in the following error:

```console
$ aiida-pseudo install sssp --download-only
Info: downloading selected pseudo potentials archive...  [OK]
Info: downloading selected pseudo potentials metadata...  [OK]
Traceback (most recent call last):
  File "/home/mbercx/.virtualenvs/aiida-dev/bin/aiida-pseudo", line 12, in <module>
    sys.exit(cmd_root())
  File "/home/mbercx/.virtualenvs/aiida-dev/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/mbercx/.virtualenvs/aiida-dev/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/mbercx/.virtualenvs/aiida-dev/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/mbercx/.virtualenvs/aiida-dev/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/mbercx/.virtualenvs/aiida-dev/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/mbercx/.virtualenvs/aiida-dev/lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/mbercx/envs/aiida-dev/code/aiida-core/aiida/cmdline/utils/decorators.py", line 65, in wrapper
    return wrapped(*args, **kwargs)
  File "/home/mbercx/envs/aiida-dev/code/aiida-pseudo/aiida_pseudo/cli/install.py", line 190, in cmd_install_sssp
    filepath.rename(filepath_target)
  File "/usr/lib/python3.8/pathlib.py", line 1340, in rename
    self._accessor.rename(self, target)
OSError: [Errno 18] Invalid cross-device link: '/tmp/tmpj8madxti/archive.tar.gz' -> '/home/mbercx/envs/aiida-dev/tmp/archive.tar.gz'
```

Looking into it, but maybe you already know the fix.

sphuber:
ah yeah, this is because of `shutil.move` I think which does not work across different file systems in some cases. Are you running this on MacOS by any chance?

mbercx:
> ah yeah, this is because of `shutil.move` I think which does not work across different file systems in some cases. Are you running this on MacOS by any chance?

No, this is on my work station that's running Ubuntu 😬

sphuber:
Ah wait, it was the inverse. The method `os.rename` (which I think `pathlib` uses under the hood if we do `Path().rename()` fails across file systems and we should use `shutil.move`. I am not sure why your temp directory is considered a different file system though. Did you mount this in a specific way or is this the default Ubuntu `/tmp` setup? Because for me it works just fine on Ubuntu.

mbercx:
Seems my `/` and `/home` are in different partitions. They are of the same _type_ of file system though (ext4):

```console
$ df -T
Filesystem     Type      1K-blocks       Used  Available Use% Mounted on
udev           devtmpfs   32743092          0   32743092   0% /dev
tmpfs          tmpfs       6553548      20696    6532852   1% /run
/dev/sda2      ext4      229185456  161335460   56138328  75% /
tmpfs          tmpfs      32767728         20   32767708   1% /dev/shm
...
/dev/sda1      vfat         523248        264     522984   1% /boot/efi
/dev/sdb2      ext4      618222752  161654032  425095056  28% /home
/dev/sdb3      ext4     1236087656  465207444  708020592  40% /scratch
...
```

sphuber:
The type shouldn't matter. Still weird that this fails, because I have literally the same setup. Anyway, it could easily fail for other OS'as well. I have changed the implementation. Can you try again with the latest commit?

mbercx:
> Can you try again with the latest commit?

Works like a charm! Continuing with review now.

aiida_pseudo/cli/params/options.py
```diff
@@ -82,3 +82,7 @@
     show_default=True,
     help='Specify the energy unit of the cutoffs. Must be recognized by the ``UnitRegistry`` of the ``pint`` library.'
 )
+
+DOWNLOAD_ONLY = OverridableOption(
+    '--download-only', is_flag=True, help='Only download the pseudopotential files and do install as a family.'
```
mbercx, line 87: I suppose you mean that they can use `aiida-pseudo install family` after the download to install the pseudo family? I'm not sure if that is clear from this message. I'd make it something like:

```suggestion
    '--download-only', is_flag=True, help='Only download the pseudopotential files to the current working directory, without installing the pseudopotential family.'
```

aiida_pseudo/groups/family/pseudo_dojo.py
```diff
@@ -3,6 +3,7 @@
 import collections
 import json
 import os
+import pathlib
```
mbercx, line 6: In `aiida_pseudo/cli/utils.py` you only importh `Path`:

```Python
from pathlib import Path
```

Since no other methods/classes from `pathlib` are used in this module, perhaps do the same here to appease my OCD? 😁

tests/cli/test_install.py
```diff
@@ -113,7 +113,7 @@ def _run_monkeypatched_install_pseudo_dojo(options=None, raises=None):
 
 @pytest.mark.usefixtures('clear_db')
 def test_install_family(run_cli_command, get_pseudo_archive):
-    """Test `aiida-pseudo install family`."""
+    """Test ``aiida-pseudo install family``."""
```
mbercx, line 116: Myes, I approve!

mbercx (approved):
Looking bueno, @sphuber! Just a few minor comments and nods of approval, please feel free to merge at your discretion.

aiida_pseudo/groups/family/pseudo_dojo.py
sphuber: but then my OCD will be triggered:
```
import collections
import json
import os
from pathlib import Path
```
is so fugly...

mbercx: Then move it down a few lines next to

```Python
from typing import Sequence
```

? 😁

aiida_pseudo/groups/family/pseudo_dojo.py
```diff
@@ -5,6 +5,8 @@
 import os
 import re
 import warnings
+
+from pathlib import Path
 from typing import Sequence
```
mbercx, line 10: So pwetty 😍

aiida_pseudo/groups/family/pseudo_dojo.py
sphuber: Your wish is my command

PR #67, opened 2021-04-09
CLI: Add `units` option to `aiida-pseudo family cutoffs set`

### LINKED ISSUE #64: CLI: Add `units` option to `aiida-pseudo family cutoffs set`
Now that we are able to specify units for the cutoffs using the delicious `pint` package, we should extend the `aiida-pseudo family cutoffs set` command with a `--units/-s` option. Currently the user can only specify the cutoffs in the default units (eV).

mbercx:
Fixes #64 

Currently the user is not able to specify the units of the plane-wave
energy cutoffs when using the `aiida-pseudo family cutoffs set` command.
This means that the user is limited to only specifying the units in
electronvolt.

Here we add the `-u/--unit` option to the `aiida-pseudo family cutoffs set`
command to allow to user to specify the units of the cutoff values
provided in the CUTOFFS JSON file.

### DIFF
--- a/aiida_pseudo/cli/family.py
+++ b/aiida_pseudo/cli/family.py
@@ -55,8 +55,9 @@ def cmd_family_cutoffs():
 @arguments.PSEUDO_POTENTIAL_FAMILY()
 @click.argument('cutoffs', type=click.File(mode='rb'))
 @options.STRINGENCY(required=True)
+@options.UNIT()
 @decorators.with_dbenv()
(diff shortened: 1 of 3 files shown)


sphuber (approved):
Slightly surprised that you would consider GME stock as invalid, but I will let it slide

mbercx:
> Slightly surprised that you would consider GME stock as invalid, but I will let it slide

Clearly GME stock is a unit of power, not energy ;)

PR #69, opened 2021-04-10
CLI: refactor download code from automated install commands

sphuber:
This serves two purposes: it makes it easier to reuse the download code
outside of the CLI and secondly, it enables the mocking of this part of
the code, which is a heavy operation. Without this, unit testing of this
command would be prohibitively expensive as it would need to download
all files for every test.

### DIFF
--- a/aiida_pseudo/cli/install.py
+++ b/aiida_pseudo/cli/install.py
@@ -2,6 +2,7 @@
 """Command to install a pseudo potential family."""
 import json
 import os
+import pathlib
 import shutil
 import tempfile
 
@@ -11,10 +12,9 @@ from aiida.cmdline.utils import decorators, echo
 from aiida.cmdline.params import options as options_core
 from aiida.cmdline.params import types
 
+from aiida_pseudo.groups.family import PseudoDojoConfiguration, SsspConfiguration
 from .params import options
 from .root import cmd_root
-
-URL_SSSP_BASE = 'https://legacy-archive.materialscloud.org/file/2018.0001/v4/'
 
 
 @cmd_root.group('install')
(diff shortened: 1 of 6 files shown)


sphuber:
I need this for a follow up PR that adds a new flag to the automated family install commands and to test it, we need to mock the downloading because that would be too expensive.

aiida_pseudo/cli/install.py
```diff
@@ -65,6 +65,80 @@ def cmd_install_family(archive, label, description, archive_format, family_type,
     echo.echo_success(f'installed `{label}` containing {family.count()} pseudo potentials')
 
 
+def download_sssp(
+    configuration: SsspConfiguration,
+    filepath_archive: pathlib.Path,
+    filepath_metadata: pathlib.Path,
+    traceback: bool = False
+):
```
mbercx, line 73:
```suggestion
) -> None:
```

sphuber:
Also, I haven't added the monkeypatched download function for PseudoDojo yet. Will try to add that

aiida_pseudo/cli/install.py
```diff
@@ -65,6 +65,80 @@ def cmd_install_family(archive, label, description, archive_format, family_type,
...
+            handle.write(response.content)
+            handle.flush()
+
+
+def download_pseudo_dojo(
+    configuration: SsspConfiguration,
+    filepath_archive: pathlib.Path,
+    filepath_metadata: pathlib.Path,
+    traceback: bool = False
+):
```
mbercx, line 110:
```suggestion
) -> None:
```

tests/cli/test_install.py
```diff
@@ -1,11 +1,60 @@
...
+@pytest.fixture
+def run_monkeypatched_install_sssp(run_cli_command, get_pseudo_potential_data, monkeypatch, tmpdir):
+    """Fixture that monkey patches the ``aiida_pseudo.cli.install.download_sssp`` method and calls the install cmd."""
+
+    def download_sssp(
+        configuration: SsspConfiguration,
+        filepath_archive: pathlib.Path,
+        filepath_metadata: pathlib.Path,
+        traceback: bool = False
+    ):  # pylint: disable=unused-argument
```
mbercx, line 25:
```suggestion
    ) -> None:  # pylint: disable=unused-argument
```

mbercx (dismissed):
Muy bien, señor @sphuber! Just left a few comments re type hinting so you don't get the impression I'm just approving your PR's without looking at the code. 😅

sphuber:
Much gracias hefe @mbercx I have addressed your comments and I have added the monkey patched PseudoDojo download method with a simple test

mbercx (approved):
Kus van de juf en een bank vooruit!

PR #71, opened 2021-04-12
CLI: do not override existing stringencies in `family cutoffs set`

### LINKED ISSUE #66: CLI: `aiida-pseudo family cutoffs set` removes all other stringencies
Currently, the `aiida-pseudo family cutoffs set` command will take the provided `stringency` and set _only_ cutoffs for that stringency, making it the default. This means that any other stringencies will be removed from the pseudo family. I'm not sure if this is the desired behaviour, since it limits the user to only defining one stringency at a time, and could be potentially dangerous. If the user first installs a specific pseudo family with a set of recommended cutoffs, and then uses the `aiida-pseudo family cutoffs set` command when trying to add a _new_ stringency, the other recommended values will be lost.

~~Perhaps we should consider matching the `set_cutoffs` and `get_cutoffs` API's more closely. Currently their behaviour is different in the sense that the first sets _all_ cutoffs, and the second only retrieves the cutoffs for a _specified_ stringency. I must admit that this threw me off at first too. I would suggest:~~

~~* Renaming the `set_cutoffs` command to `set_cutoffs_all` or something similar.~~
~~* Adding a `set_cutoffs` command that allows the user to set the cutoffs for a single stringency.~~

~~This would also avoid awkward workarounds where in order to set a single stringency you have to get the whole dictionary with the hidden method `_get_cutoffs` (which I would consider renaming to `_get_cutoffs_dict` to make it more distinguishable from `get_cutoffs`) and then override one key and then set the cutoffs with `set_cutoffs`.~~

Second part of discussion moved to #68.

sphuber:
Fixes #66 

The command's interface only allows to specify a single stringency at a
time, however, it would use the `set_cutoffs` method internally, which
would override any existing stringencies. This bug is addressed by first
retrieving existing stringencies before adding the new one to it and
setting the ensemble as the new stringencies.

There is an additional limitation that needs to be taken into account,
however. Since the current design only allows for a single energy unit
to be used at a time for all stringencies, when adding a new stringency,
the same unit has to be used. If, in the future, `set_cutoffs` starts to
support setting individual stringencies additively and the family
supports multiple units for the different stringencies, then all of this
logic can be simplified significantly and the limitations removed.

### DIFF
--- a/aiida_pseudo/cli/family.py
+++ b/aiida_pseudo/cli/family.py
@@ -86,8 +86,17 @@ def cmd_family_cutoffs_set(family, cutoffs, stringency, unit):  # noqa: D301
     except ValueError as exception:
         raise click.BadParameter(f'`{cutoffs.name}` contains invalid JSON: {exception}', param_hint='CUTOFFS')
 
+    # This limitation can be removed once ``set_cutoffs`` allows to set additive stringencies and each stringency can
+    # define its own unit.
+    current_unit = family.get_cutoffs_unit()
+    if unit != current_unit:
+        raise click.BadParameter(f'`{unit}` does not match the unit of the family `{current_unit}`', param_hint='UNIT')
+
     try:
-        family.set_cutoffs({stringency: data}, unit=unit)
+        # This code can also be simplified once ``set_cutoffs`` allows to set individual stringencies additively.
+        current_cutoffs = family._get_cutoffs()  # pylint: disable=protected-access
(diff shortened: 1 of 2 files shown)


sphuber:
@mbercx thought to fix this bug before you start implementing the new interface. I left comments in the code where workarounds are currently added, which can be simplified when that happens.

mbercx (approved):
On the eight day he reviewed @sphuber's code, and he saw that it was good.

PR #72, opened 2021-04-12
`RecommendedCutoffMixin`: set only one stringency with `set_cutoffs`

### LINKED ISSUE #68: `RecommendedCutoffMixin`: change`set_cutoffs` API to only set one stringency
Currently the behaviour of  the `set_cutoffs` and `get_cutoffs` methods of the `RecommendedCutoffMixin` class is different. The first sets the cutoffs for all stringencies along with a default stringency, and the second only retrieves the cutoffs for a specified stringency. This can be somewhat confusing for the user, and means that the user cannot easily set the cutoffs for a single stringency.

It would be more intuitive to change the `set_cutoffs` method to only set the cutoffs for a single stringency. This way changing a single stringency is straightforward, and the user can simply call the method in a loop to set multiple stringencies. In addition, we can add the `set_default_stringency` method so the user can still specify a default.

mbercx:
Fixes #68 

Currently the behaviour of the `set_cutoffs` and `get_cutoffs` methods of the
`RecommendedCutoffMixin` class is different. The first sets the cutoffs for all
stringencies along with a default stringency, the second only retrieves the
cutoffs for a specified stringency. This can be somewhat confusing for the
user, and means that the user cannot easily set the cutoffs for a single
stringency.

Here we change the `set_cutoffs` method so it only sets the recommended
cutoffs for a single stringency, which has to be specified as an input
argument. The stringency is automatically set to be the default in case
it is the only stringency specified, and a `set_default_stringency`
method is added to the user can still change the default stringency at
any time after setting multiple stringencies.

We also allow the user to specify units for each stringency, extending
the value corresponding to the `_key_cutoffs_unit` in the extras to a
dictionary that specifies the units for each stringency, similar to the
cutoffs.

### DIFF
--- a/aiida_pseudo/groups/mixins/cutoffs.py
+++ b/aiida_pseudo/groups/mixins/cutoffs.py
@@ -30,37 +30,38 @@ class RecommendedCutoffMixin:
         """Validate a cutoff dictionary for a given set of elements.
 
         :param elements: set of elements for which to validate the cutoffs dictionary.
-        :param cutoffs: dictionary with recommended cutoffs. Format: multiple sets of recommended cutoffs can be
-            specified where the key represents the stringency, e.g. ``low`` or ``normal``. For each stringency, a
-            dictionary should be defined that for each element symbols for which the family contains a pseudopotential,
-            two values are specified, ``cutoff_wfc`` and ``cutoff_rho``, containing a float value with the recommended
-            cutoff to be used for the wave functions and charge density, respectively.
+        :param cutoffs: dictionary with recommended cutoffs. Format: set of recommended cutoffs written as a
+            dictionary that maps each element for which the family contains a pseudopotential to a dictionary that
+            specifies the ``cutoff_wfc`` and ``cutoff_rho`` keys, corresponding a float value with the recommended
+            cutoff to be used for the wave functions and charge density, respectively. For example:
+
+                {
+                    "Ag": {
+                        "cutoff_wfc": 50.0,
+                        "cutoff_rho": 200.0
+                    },
+                    ...
+                }
         :raises ValueError: if the set of elements and those defined in the cutoffs do not match exactly, or if the
             cutoffs dictionary has an invalid format.
         """
         elements_family = set(elements)
 
-        for stringency, cutoffs_sub in cutoffs.items():
+        elements_cutoffs = set(cutoffs.keys())
+        elements_diff = elements_family ^ elements_cutoffs
 
-            elements_cutoffs = set(cutoffs_sub.keys())
-            elements_diff = elements_family ^ elements_cutoffs
+        if elements_family < elements_cutoffs:
+            raise ValueError(f'cutoffs defined for unsupported elements: {elements_diff}')
 
-            if elements_family < elements_cutoffs:
-                raise ValueError(f'cutoffs defined for unsupported elements: {elements_diff}')
+        if elements_family > elements_cutoffs:
+            raise ValueError(f'cutoffs not defined for all family elements: {elements_diff}')
 
-            if elements_family > elements_cutoffs:
-                raise ValueError(f'cutoffs not defined for all family elements: {elements_diff}')
+        for element, values in cutoffs.items():
+            if set(values.keys()) != {'cutoff_wfc', 'cutoff_rho'}:
+                raise ValueError(f'invalid cutoff keys for element {element}: {values}')
 
-            for element, values in cutoffs_sub.items():
-                if set(values.keys()) != {'cutoff_wfc', 'cutoff_rho'}:
-                    raise ValueError(
-                        f'invalid cutoff keys for stringency `{stringency}` and element {element}: {values}'
-                    )
-
-                if any(not isinstance(cutoff, (int, float)) for cutoff in values.values()):
-                    raise ValueError(
-                        f'invalid cutoff values for stringency `{stringency}` and element {element}: {values}'
-                    )
+            if any(not isinstance(cutoff, (int, float)) for cutoff in values.values()):
+                raise ValueError(f'invalid cutoff values for element {element}: {values}')
 
     @staticmethod
     def validate_cutoffs_unit(unit: str) -> None:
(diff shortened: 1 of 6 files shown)


mbercx:
@sphuber test should be passing, but could still do with a good scrubbin'. Will do so once we are agreed on the design!

aiida_pseudo/groups/mixins/cutoffs.py
```diff
@@ -90,60 +91,92 @@ def validate_stringency(self, stringency: str) -> None:
...
-    def _get_cutoffs(self) -> dict:
-        """Return the cutoffs dictionary.
+    def _get_cutoffs_dict(self) -> dict:
+        """Return the cutoffs dictionary that maps the stringencies to the recommended cutoffs.
 
         :return: the cutoffs extra or an empty dictionary if it has not yet been set.
         """
         return self.get_extra(self._key_cutoffs, {})
 
+    def _get_cutoffs_unit_dict(self) -> str:
```
sphuber, line 101:
```suggestion
    def _get_cutoffs_unit_dict(self) -> dict:
```

aiida_pseudo/groups/mixins/cutoffs.py
```diff
@@ -90,60 +91,92 @@ def validate_stringency(self, stringency: str) -> None:
...
+        :return: the cutoffs units extra or an empty dictionary if it has not yet been set.
+        """
+        return self.get_extra(self._key_cutoffs_unit, {})
+
     def get_default_stringency(self) -> Union[str, None]:
         """Return the default stringency if defined.
 
         :return: the default stringency or ``None`` if no cutoffs have been defined for this family.
-        :raises ValueError: if not default stringency has been defined.
+        :raises ValueError: if default stringency has not been defined.
```
sphuber, line 112: I noticed that the current docstring and return type of `get_default_stringency` is incorrect. It says it returns `None` if not defined but it actually raises. Maybe raising is actually the best behavior? If we have it return `None`, we should maybe change the implementation of `get_cutoffs_unit` which now assumes it gets a value back

aiida_pseudo/groups/mixins/cutoffs.py
```diff
@@ -90,60 +91,92 @@ def validate_stringency(self, stringency: str) -> None:
...
         """
         try:
             return self.get_extra(self._key_default_stringency)
         except AttributeError as exception:
             raise ValueError('no default stringency has been defined.') from exception
 
+    def set_default_stringency(self, default_stringency: str) -> None:
+        """Set the default stringency for the recommended cutoffs.
+
+        :param default_stringency: the default stringency to be used when ``get_recommended_cutoffs`` is called. If is
```
sphuber, line 122:
```suggestion
        :param default_stringency: the default stringency to be used when ``get_recommended_cutoffs`` is called. It is
```

aiida_pseudo/groups/mixins/cutoffs.py
```diff
@@ -90,60 +91,92 @@ def validate_stringency(self, stringency: str) -> None:
...
+            possible to not specify this if and only if the cutoffs only contain a single stringency set. That one will
+            then automatically be set as default.
```
sphuber, line 124: I think these last two sentences of the docstring actually apply to `set_cutoffs` doesn't it? Here it doesn't make much sense.

aiida_pseudo/groups/mixins/cutoffs.py
```diff
@@ -90,60 +91,92 @@ def validate_stringency(self, stringency: str) -> None:
...
+        :raises ValueError: if the provided default stringency is not in the tuple of available cutoff stringencies for
+            the pseudo family.
+        """
+        if default_stringency not in self.get_cutoff_stringencies():
+            raise ValueError('provided default stringency not in tuple of available cutoff stringencies.')
```
sphuber, line 129:
```suggestion
            raise ValueError(f'provided default stringency not in tuple of available cutoff stringencies: {self.get_cutoff_stringencies}.')
```

aiida_pseudo/groups/mixins/cutoffs.py
```diff
@@ -90,60 +91,92 @@ def validate_stringency(self, stringency: str) -> None:
...
 
-    def set_cutoffs(self, cutoffs: dict, default_stringency: str = None, unit: str = None) -> None:
-        """Set the recommended cutoffs for the pseudos in this family.
+    def set_cutoffs(self, cutoffs: dict, stringency: str, unit: str = None) -> None:
+        """Set the recommended cutoffs for the pseudos in this family and a specified stringency.
 
-        .. note:: units of the cutoffs should be in electronvolt.
+        .. note: If there is only one stringency defined for the pseudo family, this method will automatically set this
+            as the default. Use the ``set_default_stringency`` method to change the default when setting multiple
+            stringencies.
```
sphuber, line 145:
```suggestion
        .. note: If, after the cutoffs have been set, there is only one stringency defined for the pseudo family, this method will automatically set this
            as the default. Use the ``set_default_stringency`` method to change the default when setting multiple
            stringencies.
```

aiida_pseudo/groups/mixins/cutoffs.py
```diff
@@ -155,7 +188,19 @@ def get_cutoffs(self, stringency=None) -> Union[dict, None]:
         stringency = stringency or self.get_default_stringency()
 
         try:
-            return self._get_cutoffs()[stringency]
+            return self._get_cutoffs_dict()[stringency]
```
sphuber, line 191: Should probably add to the docstring:
```
:raises ValueError: if no stringency is specified and no default stringency is defined for the family.
```

aiida_pseudo/groups/mixins/cutoffs.py
```diff
@@ -155,7 +188,19 @@ def get_cutoffs(self, stringency=None) -> Union[dict, None]:
...
+        except KeyError as exception:
+            raise ValueError(f'stringency `{stringency}` is not defined for this family.') from exception
+
+    def get_cutoffs_unit(self, stringency: str = None) -> str:
+        """Return the cutoffs unit.
```
sphuber, line 196:
```suggestion
        """Return the cutoffs unit for the specified or family default stringency.
```

aiida_pseudo/groups/mixins/cutoffs.py
```diff
@@ -155,7 +188,19 @@ def get_cutoffs(self, stringency=None) -> Union[dict, None]:
...
+
```
sphuber, line 197:
```suggestion
:param stringency: optional stringency for which to retrieve the unit. If not specified, the default stringency of the family is uses.
```

tests/cli/test_family.py
```diff
@@ -13,32 +13,49 @@
 
 
 @pytest.fixture
-def generate_cutoffs(tmp_path):
+def generate_cutoffs_dict(tmp_path):
```
sphuber, line 16:
```suggestion
def generate_cutoffs_dict():
```

tests/cli/test_family.py
```diff
@@ -13,32 +13,49 @@
...
-                cutoffs[stringency][element] = {'cutoff_wfc': 1.0, 'cutoff_rho': 2.0}
+                cutoffs_dict[stringency][element] = {'cutoff_wfc': 1.0, 'cutoff_rho': 2.0}
+
+        return cutoffs_dict
+
+    return _generate_cutoffs_dict
+
+
+@pytest.fixture
+def generate_cutoffs(tmp_path):
```
sphuber, line 34:
```suggestion
def generate_cutoffs():
```

tests/cli/test_family.py
sphuber: You can also have this fixture reuse the `generate_cutoffs` fixture perhaps to generate the inner structure?

tests/cli/test_family.py
```diff
@@ -57,28 +74,28 @@ def test_family_cutoffs_set(run_cli_command, get_pseudo_family, generate_cutoffs
...
     family = get_pseudo_family(cls=CutoffsFamily)
     stringency = 'normal'
-    cutoffs = generate_cutoffs(family, stringencies=(stringency,))
+    cutoffs = generate_cutoffs(family)
 
     # Currently, the CLI checks that the unit set matches the one already set on the family, because only a single
     # unit can be used at a time. This limitation will soon be removed though, at which point, this family should have
     # eV as a unit, such that the test of the CLI call to set it to hartree actually tests that the unit is changed.
-    family.set_cutoffs(cutoffs, unit='hartree', default_stringency=stringency)
+    family.set_cutoffs(cutoffs, stringency, 'hartree')
```
sphuber, line 95:
```suggestion
    family.set_cutoffs(cutoffs, stringency, 'eV')
```
use an explicitly different unit than the one we set later for the actual test. This is now possible because we allow multiple units to be specified. You can therefore also remove the three comments above this, or replace them with an explanation why we set `eV` explicitly (this is just in case in the future for some reason the default unit is changed to `hartree`)

sphuber (changes requested):
Thanks @mbercx some minor adjustments but I think the design of the interface is good like this. Maybe just add some tests surrounding the fact that we now have multiple units and add a test on what should happen if you call it and no default stringency is set.

aiida_pseudo/groups/mixins/cutoffs.py
mbercx: I guess it depends on what error we want the user to see when they try to use `get` methods and have no stringencies defined. Once _any_ stringency is defined with the `set_cutoffs` method, it will automatically be set to the default. I think raising with an error message like `'no recommended cutoffs have been specified for this pseudo family'` might be better than `'no default stringency has been defined.'` when no stringencies have been defined _at all_.

However, now that I think of it, we should probably also add a `remove_cutoffs` method, since now there is no API for doing so. If this method is implemented, it's possible that the user:

1. Uses `set_cutoffs` to set the `'normal'` stringency, which becomes the default.
2. Uses `set_cutoffs` to set to `'high'` stringency.
3. Removes the `'normal'` stringency.

What do we do in this instance? Since there is only one stringency left, it makes sense to make that the default, but what if there are 3 stringencies at some point, and the default is removed? Do we let the user simply not remove the default stringency (raise), or allow stringencies/cutoffs to be defined without having a default set?

aiida_pseudo/groups/mixins/cutoffs.py
mbercx: Oops! Copy 🍝 ! 😬

aiida_pseudo/groups/mixins/cutoffs.py
sphuber: Good points. Seems like that really we should just accept that it is very much possible to not have a default stringency defined, which means that we should simply have it raise when it is asked for. I think that `'no default stringency has been defined'` in this case is then better than `'no recommended cutoffs have been specified for this pseudo family' ` because not only does the latter tie an implementation detail into error message, but once we add the `delete_cutoffs` method, that won't necessarily be true anymore.

TLDR: my vote
* let `get_default_stringency` raise if not defined
* `set_cutoffs` automatically sets default if only one stringency is defined
* implement `delete_cutoffs` that will also remove the default stringency if that corresponds (only sets new default if only stringency remains)

tests/cli/test_family.py
mbercx: Ah, I missed that comment while swashbuckling through the tests somehow. My bad!

aiida_pseudo/groups/mixins/cutoffs.py
```diff
@@ -90,72 +89,143 @@ def validate_stringency(self, stringency: str) -> None:
...
+        :raises ValueError: if the requested stringency is not defined for this family.
+        """
+        stringency = stringency or self.get_default_stringency()
+
+        try:
+            return self._get_cutoffs_dict()[stringency]
+        except KeyError as exception:
+            raise ValueError(f'stringency `{stringency}` is not defined for this family.') from exception
+
+    def delete_cutoffs(self, stringency) -> None:
```
sphuber, line 195: oliver

And maybe we do the same as in `set_cutoffs`, if after the operation is completed there is only one stringency left, that becomes the new default. If you agree, would be good to have a test for this and add in the docstring like for `set_cutoffs`

mbercx: I agree it makes sense, and don't immediately see any issue with this. I wonder if we should print a message when the user removes the `default_stringency` with the `remove_cutoffs` method. Something along the lines of:

* Only one stringency remaining: `The default stringency has been set to: '<LEFTOVER STRINGENCY>'`
* Multiple stringencies remaining: `Default stringency removed but multiple options left. Please set the new default stringency with the ``set_default_stringency()`` method.`

And I will make sure that Oliver's hunger for tests is amply satisfied. 😁

sphuber: Would be fine with me, but what are you suggesting to use for printing the message? I think the only options we have are logging or a `warnings.warn` and probably the latter since we haven't really used or setup logging yet. Printing directly I don't like in normal API code tbh.

mbercx:
That should turn Oliver's frown upside down! I also implemented the logic discussed and added some warnings in case the default stringency is removed, or a new default stringency needs to still be set.

sphuber (approved):
"Do you know what I consider the greatest sin in the world, my dear? Ingratitude." But I, for one, am very grateful for this PR

PR #73, opened 2021-04-13
Dependencies: remove temporary upper limit for `sqlalchemy`

sphuber:
This reverts `383ab247c40ada60c22ce3673cbbc3efccd21f98`. The upper limit
is no longer necessary as the incompatibility of `sqlalchemy-utils` with
`sqlalchemy==1.4` has been solved downstream.

### DIFF
--- a/setup.json
+++ b/setup.json
@@ -41,8 +41,7 @@
         "click-completion~=0.5",
         "pint~=0.16.1",
         "psycopg2-binary<2.9",
-        "requests~=2.20",
-        "sqlalchemy<1.4"
+        "requests~=2.20"
     ],
     "extras_require": {
         "pre-commit": [


sphuber:
This is still failing because the limit in `aiida-core` has not been released yet. We have two options here:

1) Once `aiida-core==1.6.2` is released with the upped requirement of `sqlalchemy-utils`, we depend on that version of `aiida-core` and we remove the dependency limit on `sqlalchemy` here or
2) We simply remove the `sqlalchemy` limit here now and add a minimum requirement on `sqlalchemy-utils`.

Writing it out like this, it seems clear that 2) is the much favored option as we can keep supporting more versions of `aiida-core`.

sphuber:
Just released `aiida-core==1.5.3` which fixes the compatibility issues with `sqlalchemy` and `psycopg2-binary`, so this can now be merged.

PR #74, opened 2021-04-13
Release `v0.6.0`

sphuber:


mbercx:
Note that we should probably add a warning that these changes are quite breaking (although it can be expected since it's pre-v1.0.0). I was doing a bit of field testing to see what will happen if you try to use some of the new features with pseudos installed with 0.5.0:

```Python
In [1]: ps = load_group('SSSP/1.0/PBE/efficiency')

In [2]: ps.get_recommended_cutoffs(elements=('Li', 'Co'))
Out[2]: (612.2561276385, 4898.049021108)

In [3]: ps.get_cutoffs_unit()
KeyError                                  Traceback (most recent call last)
~/envs/aiida-dev/code/aiida-pseudo/aiida_pseudo/groups/mixins/cutoffs.py in get_cutoffs_unit(self, stringency)
    255         try:
--> 256             return self._get_cutoffs_unit_dict()[stringency]
    257         except KeyError as exception:

KeyError: 'normal'

The above exception was the direct cause of the following exception:

ValueError                                Traceback (most recent call last)
<ipython-input-3-fa1e1bfdf5b8> in <module>
----> 1 ps.get_cutoffs_unit()

~/envs/aiida-dev/code/aiida-pseudo/aiida_pseudo/groups/mixins/cutoffs.py in get_cutoffs_unit(self, stringency)
    256             return self._get_cutoffs_unit_dict()[stringency]
    257         except KeyError as exception:
--> 258             raise ValueError(f'stringency `{stringency}` is not defined for this family.') from exception
    259 
    260     def get_recommended_cutoffs(self, *, elements=None, structure=None, stringency=None, unit=None):

ValueError: stringency `normal` is not defined for this family.
```

In fact, maybe we should still fix this, i.e. have the default unit assigned for any stringency which _does_ have the cutoffs set, but _doesn't_ have the unit in the `'_cutoffs_unit'` dictionary stored in the extras. If we make sure the pseudos installed with v0.5.0 still work as intended with all features in v0.6.0, we also don't have to build in any workarounds in the plugins (see https://github.com/aiidateam/aiida-quantumespresso/issues/670).

sphuber:
If we can, in a reasonable way, build it such that existing families continue to function then I am all for it. I just would like to prevent nasty workarounds that will linger in the code for a long time if not forever. Can we summarize what the breaking changes are exactly? Is it only the change of adding multiple units?

mbercx:
> I I just would like to prevent nasty workarounds that will linger in the code for a long time if not forever. 

Yeah, I understand and agree.

> Can we summarize what the breaking changes are exactly? Is it only the change of adding multiple units?

I'd have to do a bit of field testing. Let me set up a branch and see how many workarounds we'd have to implement, and assess their level of nastiness.

mbercx:
See #75.

CHANGELOG.md
```diff
@@ -1,5 +1,32 @@
 # Change log
 
+## 0.6.0
+
+The biggest change in this release is the addition of explicit units for the recommended cutoffs of pseudopotential families that support it.
+Before, the unit always had to be specified in electronvolt, but now it is possible to define cutoffs in a variety of energy units.
+To facilitate that a single family can specify multiple sets off cutoffs at different "stringencies", with optionally different units, the interface of `RecommendedCutoffMixin.set_cutoffs` had to be changed.
```
mbercx, line 7:
```suggestion
To facilitate that a single family can specify multiple sets of cutoffs at different "stringencies", with optionally different units, the interface of `RecommendedCutoffMixin.set_cutoffs` had to be changed.
```

CHANGELOG.md
```diff
@@ -1,5 +1,32 @@
 # Change log
 
+## 0.6.0
+
+The biggest change in this release is the addition of explicit units for the recommended cutoffs of pseudopotential families that support it.
+Before, the unit always had to be specified in electronvolt, but now it is possible to define cutoffs in a variety of energy units.
```
mbercx, line 6:
```suggestion
Before, the cutoffs always had to be specified in electronvolt, but now it is possible to define cutoffs in a variety of energy units.
```

CHANGELOG.md
```diff
@@ -1,5 +1,32 @@
...
+To facilitate that a single family can specify multiple sets off cutoffs at different "stringencies", with optionally different units, the interface of `RecommendedCutoffMixin.set_cutoffs` had to be changed.
+Instead of taking a dictionary with the cutoffs for all stringencies in one go, it now takes the dictionary of cutoffs for a single stringency.
+Pseudopotential families that were installed with `aiida-pseudo==0.5.0` should continue to function properly.
```
mbercx, line 9: I think also `aiida-pseudo==0.4.0`?

mbercx (approved):
Just a few minor comments, then she's ready for launch, capt'n!

mbercx (approved):
LGTM! 🚀

PR #75, opened 2021-04-13
`RecommendedCutoffMixin`: Add support for `aiida-pseudo>=0.4.0` families

mbercx:
Currently, pseudo families installed with `aiida-pseudo>=0.4.0` will
fail to work with the new API features in two ways, both related to the
addition of units for the recommended cutoffs:

* `delete_cutoffs` expects the units to be stored in the
`'_cutoffs_unit'` extras dictionary. Simply adding `None` as a default
to return in the `.pop()` call fixes this issue.
* `get_cutoffs_unit` will return a `KeyError` since the previously
configured stringencies did not include units. Here we can add an `if`
clause that check if the requested stringency _does_ have its cutoffs
set. Since the default unit was and is `eV`, we set this as the new unit
for the stringency. From then on the pseudo family will be "fixed".

Note: families installed with `aiida-pseudo<=0.3.0` will still cause
problems since the units for the recommended cutoffs were set to
eV in v0.4.0.

### DIFF
--- a/aiida_pseudo/groups/mixins/cutoffs.py
+++ b/aiida_pseudo/groups/mixins/cutoffs.py
@@ -257,8 +257,10 @@ class RecommendedCutoffMixin:
         except KeyError as exception:
             # Workaround to deal with pseudo families installed in v0.5.0 - Set default unit in case it is not in extras
             if stringency in self.get_cutoff_stringencies():
-                self.set_cutoffs(self.get_cutoffs(stringency), stringency, unit=self.DEFAULT_UNIT)
-                return self.DEFAULT_UNIT
+                cutoffs_unit_dict = self._get_cutoffs_unit_dict()
+                cutoffs_unit_dict[stringency] = 'eV'
+                self.set_extra(self._key_cutoffs_unit, cutoffs_unit_dict)
+                return 'eV'
             # End of workaround
             raise ValueError(f'stringency `{stringency}` is not defined for this family.') from exception
 
--- a/aiida_pseudo/groups/mixins/cutoffs.py
+++ b/aiida_pseudo/groups/mixins/cutoffs.py
@@ -211,7 +211,7 @@ class RecommendedCutoffMixin:
         cutoffs_dict.pop(stringency)
 
         cutoffs_unit_dict = self._get_cutoffs_unit_dict()
-        cutoffs_unit_dict.pop(stringency)
+        cutoffs_unit_dict.pop(stringency, None)  # `None` is added to support pseudo families installed with v0.5.0
 
         self.set_extra(self._key_cutoffs, cutoffs_dict)
         self.set_extra(self._key_cutoffs_unit, cutoffs_unit_dict)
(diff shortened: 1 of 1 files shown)


mbercx:
Note that these workaround can be temporary, since if the user actually uses the `get_cutoffs_unit` method at some point (directly or indirectly), the default unit (eV) will be added to the pseudo family.

mbercx:
@sphuber If you agree with adding these workarounds, I'll also open an issue on removing them so we don't forget.

aiida_pseudo/groups/mixins/cutoffs.py
```diff
@@ -255,6 +255,11 @@ def get_cutoffs_unit(self, stringency: str = None) -> str:
         try:
             return self._get_cutoffs_unit_dict()[stringency]
         except KeyError as exception:
+            # Workaround to deal with pseudo families installed in v0.5.0 - Set default unit in case it is not in extras
+            if stringency in self.get_cutoff_stringencies():
+                self.set_cutoffs(self.get_cutoffs(stringency), stringency, unit=self.DEFAULT_UNIT)
```
sphuber, line 260: The cutoffs should already be set here and we are merely adding the unit, which should be `eV`. I would make this more explicit and actually hardcode `eV` instead of `self.DEFAULT_UNIT` in the case this changes in the future. Additionally, maybe we should just do
```
cutoffs_unit_dict = self._get_cutoffs_unit_dict()
cutoffs_unit_dict[stringency] = 'eV'
self.set_extra(self._key_cutoffs_unit, cutoffs_unit_dict)
```
It is maybe more lines, but it is showing more explicitly what we are doing. We don't need to set the cutoffs, just the unit, which was missing

aiida_pseudo/groups/mixins/cutoffs.py
```diff
@@ -255,6 +255,11 @@ def get_cutoffs_unit(self, stringency: str = None) -> str:
...
+                return self.DEFAULT_UNIT
```
sphuber, line 261:
```suggestion
                return 'eV'
```

sphuber (changes requested):
Thanks a lot @mbercx . I looked a bit through the rest of the changes in the upcoming release and I think  this should be mostly it. Just have a suggestion on implementation to make it more explicit and future proof.

sphuber:
> If you agree with adding these workarounds, I'll also open an issue on removing them so we don't forget.

Good to add an issue, but the tricky part will be deciding `when` we can remove it. Consider the example where someone is running `aiida-pseudo==0.5.0` now and installed some families. They work with this version and not until a long time do they decide to upgrade and most likely will upgrade to the latest version. If that version happens to have removed the workaround, they will run into the exception. This shows that in principle we can never know when it is safe to remove them since we don't know when "everyone" that was on v0.5 will have upgraded to a newer version with the implicit data migration.

sphuber:
> Indeed. v1.0.0 seems like a reasonable choice, but that might still be a long ways off. Still, better to have a bit of nastiness for some time that fixes the issue for most people in our massive user base ;).

Haha, still I wouldn't underestimate it. "Everyone" running `aiida-quantumespresso` and `aiida-abinit` will be affected and especially for QE there are now quite some users. Anyway, I think this workaround is not that invasive and is more than fine to keep in given the benefits it provides. No rush to get it out either.

Final thing, you verified that this then works with existing installations of `aiida-quantumespresso`? If so, I will approve merge and update the release. I think we are then ready for a release, arent we?

mbercx:
> Final thing, you verified that this then works with existing installations of `aiida-quantumespresso`? If so, I will approve merge and update the release. I think we are then ready for a release, arent we?

I opened a PR for providing support for `aiida-pseudo==0.6.0` in `aiida-quantumespresso`:

https://github.com/aiidateam/aiida-quantumespresso/pull/671

And tested the `get_builder_from_protocol()` method using both pseudos installed with `aiida-pseudo` v0.5.0 and this branch (reverting back to this branch after installing with v0.5.0). Both tests worked, so I think we're good to go!

sphuber (approved):
Love me a good ol' workaround for backwards compatibility

Issue #78, opened 2021-04-15
Add support for SG15 pseudopotentials


mbercx:
The Yambo code requires a `pw.x` calculation as a starting point for the GW calculations. However, their code currently only supports Norm-conserving pseudos, and non-linear core corrections (included in the Pseudo-dojo pseudos) are not fully implemented. Hence, @mikibonacci has requested support for the SG15 pseudos, since he would like to use the protocol methods implemented in `aiida-quantumespresso` with this family of pseudopotentials.

A few topics of discussion:

- [ ] Do we keep extending the `aiida-pseudo install` command when providing support for new pseudopotential families? 
- [ ] AFAIK, there is no detailed study that provides us with a set of recommended cutoffs for the SG15 pseudos. However, @mikibonacci suggested that perhaps the pseudo-dojo cutoffs might be a good guesstimate.

sphuber:
If I understand correctly, SG15 is not a particular format, but actual pseudo potentials. The [SSSP actually includes a number of SG15 pseudos](https://www.materialscloud.org/discover/sssp/table/efficiency). I think as a rule, we only consider adding an automated entry in `aiida-pseudo install` if the pseudo family is published, rigorously tested and provided through some persistent identifier, or at least a very stable location to download it from. People are free to create custom pseudopotential families. One can already create an SG15 family with recommended cutoffs with the current API and even CLI. All we have to do here is maybe improve the documentation on how to do this. But all the necessary functionality is there.

mbercx:
> I think as a rule, we only consider adding an automated entry in `aiida-pseudo install` if the pseudo family is published, rigorously tested and provided through some persistent identifier, or at least a very stable location to download it from. 

Sounds reasonable. Does "rigorously tested" also mean that one or more stringencies with recommended cutoffs have been identified, as was done for the SSSP? Note that there are several publications on the SG15 pseudos:

http://www.quantum-simulation.org/potentials/sg15_oncv/

You can also download the full set in XML and UPF format from the site.

> All we have to do here is maybe improve the documentation on how to do this. 

Indeed, I suppose at some point we'll have to set up proper documentation on readthedocs 😅 - Would be happy to work on this, might be a good exercise for me to set this up from scratch. Will already prepare a set of instruction for @mikibonacci for the SG15 pseudos as a starting point.

sphuber:
> Does "rigorously tested" also mean that one or more stringencies with recommended cutoffs have been identified, as was done for the SSSP

[...]

PR #80, opened 2021-04-29
Feature/install from folder

### LINKED ISSUE #79: Install family from a local folder through CLI
Currently the CLI command `aiida-pseudo install family` accepts only archives (bztar, gztar, tar, xztar, zip) as valid input arguments (either on the local file system or from URL).
It would be nice to allow also the option to pass a standard folder of the local file system. This shouldn't be difficult to implement since the underline method called by the command `aiida-pseudo install family` is `PseudoPotentialFamily.create_from_folder` that already accepts directory paths. We should also considering that old commands (like `verdi data upf uploadfamily`) were supporting this option and therefore a user starting to use the `aiida-pseudo` commands would expect a similar behavior.
In any case, even if this request won't be accepted, I would highly suggest to implement something that improves the error message of the command in the case a repository is passed. In fact, at the moment, one ends up with a `raise ValueError("unknown url type: %r" % self.full_url)` that is not very clear.

### LINKED ISSUE #81: CLI command `install family` should allow to set `pseudo_type` for the family.
Ultimately, the CLI command `install family` has been created to facilitate the access to the method `create_from_folder` of `PseudoPotentialFamily` (or its child classes) however it misses to expose an important argument of `create_from_folder`: the `pseudo_type`.
This means that currently we can create through CLI only pseudo families containing pseudos that are instances of the general pseudo class `PseudoPotentialData`. This is very limiting since most of the times the user requires to create a family containing particular kinds of pseudos (`PsmlData`,`UpfData`, etc). In fact most of the calculators (`CalJob`) accepts in input only particular kind of pseudos and never the general  `PseudoPotentialData`.

bosonie:
Fixes #79
Fixes #81

### DIFF
--- a/aiida_pseudo/cli/install.py
+++ b/aiida_pseudo/cli/install.py
@@ -5,6 +5,7 @@ import os
 import pathlib
 import shutil
 import tempfile
+import urllib.request
 
 import click
 
@@ -23,43 +24,55 @@ def cmd_install():
 
 
 @cmd_install.command('family')
-@click.argument('archive', type=types.FileOrUrl(mode='rb'))
+@click.argument('archive', type=types.PathOrUrl(exists=True, file_okay=True))
 @click.argument('label', type=click.STRING)
 @options_core.DESCRIPTION(help='Description for the family.')
 @options.ARCHIVE_FORMAT()
 @options.FAMILY_TYPE()
+@options.PSEUDO_TYPE()
 @options.TRACEBACK()
 @decorators.with_dbenv()
-def cmd_install_family(archive, label, description, archive_format, family_type, traceback):  # pylint: disable=too-many-arguments
-    """Install a standard pseudo potential family from an ARCHIVE on the local file system or from a URL.
+def cmd_install_family(archive, label, description, archive_format, family_type, pseudo_type, traceback):  # pylint: disable=too-many-arguments
+    """Install a standard pseudo potential family from a FOLDER or an ARCHIVE (on the local file system or from a URL).
 
-    The command will attempt to infer the archive format from the filename extension of the ARCHIVE. If this fails, the
-    archive format can be specified explicitly with the archive format option, which will also display which formats
-    are supported.
+    The command will attempt first to recognize the passed ARCHIVE_FOLDER as a folder in the local system. If not,
+    `archive` is assumed to be an archive and the command will attempt to infer the archive format from the
+    filename extension of the ARCHIVE. If this fails, the archive format can be specified explicitly with the archive
+    format option, which will also display which formats are supported.
 
-    By default, the command will create a base `PseudoPotentialFamily`, but the type can be changed with the family
+    By default, the command will create a base `PseudoPotentialFamily`, but the type can be changed with the pseudos
     type option. If the base type is used, the pseudo potential files in the archive *have* to have filenames that
     strictly follow the format `ELEMENT.EXTENSION`, because otherwise the element cannot be determined automatically.
     """
     from .utils import attempt, create_family_from_archive
 
-    # The `archive` is now either a `http.client.HTTPResponse` or a normal filelike object, so we get the original file
-    # name in a different way.
-    try:
-        suffix = os.path.basename(archive.url)
-    except AttributeError:
-        suffix = os.path.basename(archive.name)
+    # `archive` is a simple string, containing the name of the folder / file / url location.
 
-    # Write the content of the archive to a temporary file, because `create_family_from_archive` does currently not
-    # accept filelike objects because the underlying `shutil.unpack_archive` does not. Likewise, `unpack_archive` will
-    # attempt to deduce the archive format from the filename extension, so it is important we maintain the original
-    # filename. Of course if this fails, users can specify the archive format explicitly wiht the corresponding option.
-    with tempfile.NamedTemporaryFile(mode='w+b', suffix=suffix) as handle:
-        shutil.copyfileobj(archive, handle)
-        handle.flush()
-
+    if pathlib.Path(archive).is_dir():
+        try:
+            family = family_type.create_from_folder(archive, label, pseudo_type=pseudo_type)
+        except ValueError as exception:
+            raise OSError(f'failed to parse pseudos from `{archive}`: {exception}') from exception
+    elif pathlib.Path(archive).is_file():
         with attempt('unpacking archive and parsing pseudos... ', include_traceback=traceback):
-            family = create_family_from_archive(family_type, label, pathlib.Path(handle.name), fmt=archive_format)
+            family = create_family_from_archive(
+                family_type, label, pathlib.Path(archive), fmt=archive_format, pseudo_type=pseudo_type
+            )
+    else:
+        # The file of the url must be copied to a local temporary file. Maybe better ways to do it?
+        # The `create_family_from_archive` does currently not accept filelike objects because the underlying
+        # `shutil.unpack_archive` does not. Likewise, `unpack_archive` will attempt to deduce the archive format
+        # from the filename extension, so it is important we maintain the original filename.
+        # Of course if this fails, users can specify the archive format explicitly wiht the corresponding option.
+        with urllib.request.urlopen(archive) as handle:
+            suffix = os.path.basename(handle.url)
+            with tempfile.NamedTemporaryFile(mode='w+b', suffix=suffix) as handle:
+                shutil.copyfileobj(handle, handle)
+                handle.flush()
+                with attempt('unpacking archive and parsing pseudos... ', include_traceback=traceback):
+                    family = create_family_from_archive(
+                        family_type, label, pathlib.Path(handle.name), fmt=archive_format, pseudo_type=pseudo_type
+                    )
 
     family.description = description
     echo.echo_success(f'installed `{label}` containing {family.count()} pseudo potentials')
--- a/aiida_pseudo/cli/install.py
+++ b/aiida_pseudo/cli/install.py
@@ -1,20 +1,18 @@
 # -*- coding: utf-8 -*-
 """Command to install a pseudo potential family."""
 import json
-import os
 import pathlib
 import shutil
 import tempfile
-import urllib.request
 
 import click
+import requests
 
 from aiida.cmdline.utils import decorators, echo
 from aiida.cmdline.params import options as options_core
-from aiida.cmdline.params import types
 
 from aiida_pseudo.groups.family import PseudoDojoConfiguration, SsspConfiguration
-from .params import options
+from .params import options, types
 from .root import cmd_root
 
 
@@ -28,54 +26,70 @@ def cmd_install():
 @click.argument('label', type=click.STRING)
 @options_core.DESCRIPTION(help='Description for the family.')
 @options.ARCHIVE_FORMAT()
-@options.FAMILY_TYPE()
+@options.FAMILY_TYPE(
+    type=types.PseudoPotentialFamilyTypeParam(exclude=('pseudo.family.sssp', 'pseudo.family.pseudo_dojo'))
+)
 @options.PSEUDO_TYPE()
 @options.TRACEBACK()
 @decorators.with_dbenv()
 def cmd_install_family(archive, label, description, archive_format, family_type, pseudo_type, traceback):  # pylint: disable=too-many-arguments
-    """Install a standard pseudo potential family from a FOLDER or an ARCHIVE (on the local file system or from a URL).
+    """Install a standard pseudopotential family from an ARCHIVE.
 
-    The command will attempt first to recognize the passed ARCHIVE_FOLDER as a folder in the local system. If not,
-    `archive` is assumed to be an archive and the command will attempt to infer the archive format from the
+    The ARCHIVE can be a (compressed) archive of a directory containing the pseudopotentials on the local file system or
+    provided by an HTTP URL. Alternatively, it can be a normal directory on the local file system. The (unarchived)
+    directory should only contain the pseudopotential files and they cannot be in any subdirectory. In addition,
+    depending on the chosen pseudopotential type (see the option `-P/--pseudo-type`) there can be additional
+    requirements on the pseudopotential file and filename format.
+
+    If the ARCHIVE corresponds to a (compressed) archive, the command will attempt to infer the archive format from the
     filename extension of the ARCHIVE. If this fails, the archive format can be specified explicitly with the archive
-    format option, which will also display which formats are supported.
+    format option `-f/--archive-format`, which will also display which formats are supported. These format suffixes
+    follow the naming of the `shutil.unpack_archive` standard library method.
 
-    By default, the command will create a base `PseudoPotentialFamily`, but the type can be changed with the pseudos
-    type option. If the base type is used, the pseudo potential files in the archive *have* to have filenames that
-    strictly follow the format `ELEMENT.EXTENSION`, because otherwise the element cannot be determined automatically.
+    Once the ARCHIVE is downloaded, uncompressed and unarchived into a directory on the local file system, the command
+    will create a `PseudoPotentialFamily` instance where the type of the pseudopotential data nodes that are stored
+    within it is set through the `-P/--pseudo-type` option. If the default, `pseudo` (which corresponds to the data
+    plugin `PseudoPotentialData`), is used, the pseudopotential files in the archive *have* to have filenames that
+    strictly follow the format `ELEMENT.EXTENSION`, or the creation of the family will fail. This is because for the
+    default pseudopotential type, the format of the file is unknown and the family requires the element to be known,
+    which in this case can then only be parsed from the filename.
+
+    The pseudopotential family type that is created can also be changed with the `-F/--family-type` option. Note,
+    however, that not all values are accepted. For example, the `pseudo.family.sssp` and `pseudo.family.pseudo_dojo` are
+    blacklisted since they have their own dedicated commands in `install sssp` and `install pseudo-dojo`, respectively.
     """
     from .utils import attempt, create_family_from_archive
 
-    # `archive` is a simple string, containing the name of the folder / file / url location.
-
-    if pathlib.Path(archive).is_dir():
-        try:
+    if isinstance(archive, pathlib.Path) and archive.is_dir():
+        with attempt(f'creating a pseudopotential family from directory `{archive}`...', include_traceback=traceback):
             family = family_type.create_from_folder(archive, label, pseudo_type=pseudo_type)
-        except ValueError as exception:
-            raise OSError(f'failed to parse pseudos from `{archive}`: {exception}') from exception
-    elif pathlib.Path(archive).is_file():
+    elif isinstance(archive, pathlib.Path) and archive.is_file():
         with attempt('unpacking archive and parsing pseudos... ', include_traceback=traceback):
             family = create_family_from_archive(
-                family_type, label, pathlib.Path(archive), fmt=archive_format, pseudo_type=pseudo_type
+                family_type, label, archive, fmt=archive_format, pseudo_type=pseudo_type
             )
     else:
-        # The file of the url must be copied to a local temporary file. Maybe better ways to do it?
-        # The `create_family_from_archive` does currently not accept filelike objects because the underlying
-        # `shutil.unpack_archive` does not. Likewise, `unpack_archive` will attempt to deduce the archive format
-        # from the filename extension, so it is important we maintain the original filename.
-        # Of course if this fails, users can specify the archive format explicitly wiht the corresponding option.
-        with urllib.request.urlopen(archive) as handle:
-            suffix = os.path.basename(handle.url)
-            with tempfile.NamedTemporaryFile(mode='w+b', suffix=suffix) as handle:
-                shutil.copyfileobj(handle, handle)
-                handle.flush()
-                with attempt('unpacking archive and parsing pseudos... ', include_traceback=traceback):
-                    family = create_family_from_archive(
-                        family_type, label, pathlib.Path(handle.name), fmt=archive_format, pseudo_type=pseudo_type
-                    )
+        # At this point, we can assume that it is not a valid filepath on disk, but rather a URL and the ``archive``
+        # variable will contain the result objects from the ``requests`` library. The validation of the URL will already
+        # have been done by the ``PathOrUrl`` parameter type, so the URL is reachable. The content of the URL must be
+        # copied to a local temporary file because `create_family_from_archive` does currently not accept filelike
+        # objects, because in turn the underlying `shutil.unpack_archive` does not. In addition, `unpack_archive` will
+        # attempt to deduce the archive format from the filename extension, so it is important we maintain the original
+        # filename. Of course if this fails, users can specify the archive format explicitly with the corresponding
+        # option. We get the filename by converting the URL to a ``Path`` object and taking the filename, using that as
+        # a suffix for the temporary file that is generated on disk to copy the content to.
+        suffix = pathlib.Path(archive.url).name
+        with tempfile.NamedTemporaryFile(mode='w+b', suffix=suffix) as handle:
+            handle.write(archive.content)
+            handle.flush()
+
+            with attempt('unpacking archive and parsing pseudos... ', include_traceback=traceback):
+                family = create_family_from_archive(
+                    family_type, label, pathlib.Path(handle.name), fmt=archive_format, pseudo_type=pseudo_type
+                )
 
     family.description = description
-    echo.echo_success(f'installed `{label}` containing {family.count()} pseudo potentials')
+    echo.echo_success(f'installed `{label}` containing {family.count()} pseudopotentials')
 
 
 def download_sssp(
@@ -91,8 +105,6 @@ def download_sssp(
     :param filepath_metadata: absolute filepath to write the metadata file to.
     :param traceback: boolean, if true, print the traceback when an exception occurs.
     """
-    import requests
-
     from aiida_pseudo.groups.family import SsspFamily
     from .utils import attempt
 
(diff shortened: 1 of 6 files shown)


bosonie:
@sphuber I implemented a fix that touches the current code as little as possible. However it is worth it a more consistent refactoring, meaning modifications to the `create_family_from_archive`. What do you think?
Regarding the CI fails. It is for the pylint `consider-using-with`, however I do not think that nesting `with` commands is the way. So I leave it like that and fix it afer your opinion on my implementation.
I also fixed the CI problem every packaged had (pinned postgresql version), but if you prefer, I can remove it and do a separate PR.

bosonie:
I'm also noticing now that 
```
aiida-pseudo install family https://legacy-archive.materialscloud.org/file/2018.0001/v4/SSSP_1.0_PBE_efficiency.tar.gz label
```
that should be tested in `tests/cli/test_install.py` does actually fail in real life. Not of because of a problem in the CLI command, but for failing of the underline `family_type.create_from_folder`. Is it expected?

bosonie:
@sphuber @mbercx This is ready for review.

sphuber:
> I'm also noticing now that
> 
> ```
> aiida-pseudo install family https://legacy-archive.materialscloud.org/file/2018.0001/v4/SSSP_1.0_PBE_efficiency.tar.gz label
> ```
> 
> that should be tested in `tests/cli/test_install.py` does actually fail in real life. Not because of a problem in the CLI command, but for failing of the underline `family_type.create_from_folder`. Is it expected?

How is it failing exactly? And is it failing on the `master` branch or with your patch? Or both?

aiida_pseudo/cli/params/options.py
```diff
@@ -91,3 +91,12 @@
         'pseudopotential family.'
     )
 )
+
+PSEUDOS_TYPE = OverridableOption(
```
sphuber, line 95:
```suggestion
PSEUDO_TYPE = OverridableOption(
```

aiida_pseudo/cli/params/options.py
```diff
@@ -91,3 +91,12 @@
...
+    '-P',
+    '--pseudos-type',
```
sphuber, line 97:
```suggestion
    '--pseudo-type',
```

aiida_pseudo/cli/params/options.py
```diff
@@ -91,3 +91,12 @@
...
+    type=PseudoPotentialTypeParam(),
+    default='pseudo',
+    show_default=True,
+    help='Choose the type of pseudo potentials belonging to the family.'
```
sphuber, line 101:
```suggestion
    help='Select the pseudopotential type to be used for the family. Should be the entry point name of a subclass of `PseudoPotentialData`.'
```

aiida_pseudo/cli/params/types.py
```diff
@@ -5,7 +5,45 @@
...
+
+        try:
+            pseudo_type = DataFactory(value)
+        except exceptions.EntryPointError as exception:
+            raise click.BadParameter(f'`{value}` is not an existing data plugin.') from exception
+
+        if not issubclass(pseudo_type, PseudoPotentialData):
+            raise click.BadParameter(f'`{value}` entry point is not a subclass of `PseudoPotentialData`.')
+
+        PseudoPotentialData.entry_point = value
```
sphuber, line 35: I noticed that this line is also there for `PseudoPotentialFamilyParam` but I don't think it does anything and so is better to be removed

aiida_pseudo/cli/params/types.py
```diff
@@ -5,7 +5,45 @@
...
 from aiida.cmdline.params.types import GroupParamType
 
-__all__ = ('PseudoPotentialFamilyTypeParam', 'PseudoPotentialFamilyParam')
+__all__ = ('PseudoPotentialFamilyTypeParam', 'PseudoPotentialFamilyParam', 'PseudoPotentialTypeParam')
+
+
+class PseudoPotentialTypeParam(click.ParamType):
+    """Parameter type for `click` commands to define a subclass of `PseudoPotentialData`."""
+
+    name = 'pseudos_type_type'
```
sphuber, line 14:
```suggestion
    name = 'pseudo_type'
```

aiida_pseudo/cli/params/types.py
```diff
@@ -5,7 +5,45 @@
...
+
+    def convert(self, value, _, __):
+        """Convert the entry point name to the corresponding class.
+
+        :param value: entry point name that should correspond to subclass of `PseudoPotentialFamily` group plugin
```
sphuber, line 19:
```suggestion
        :param value: entry point name that should correspond to subclass of `PseudoPotentialData` data plugin
```

aiida_pseudo/cli/params/types.py
```diff
@@ -5,7 +5,45 @@
...
+        :return: the `PseudoPotentialFamily` subclass
```
sphuber, line 20:
```suggestion
        :return: the `PseudoPotentialData` subclass
```

aiida_pseudo/cli/params/types.py
```diff
@@ -5,7 +5,45 @@
...
+        :raises: `click.BadParameter` if the entry point cannot be loaded or is not subclass of `PseudoPotentialFamily`
```
sphuber, line 21:
```suggestion
        :raises: `click.BadParameter` if the entry point cannot be loaded or is not subclass of `PseudoPotentialData`
```

tests/cli/test_install.py
```diff
@@ -129,6 +130,26 @@ def test_install_family(run_cli_command, get_pseudo_archive):
     assert len(family.pseudos) != 0
 
 
+pytest.mark.usefixtures('clear_db')
+
+
```
sphuber, line 135:
```suggestion
```

tests/cli/test_install.py
```diff
@@ -13,6 +13,7 @@
 from aiida_pseudo.groups.family import PseudoPotentialFamily
 from aiida_pseudo.groups.family.pseudo_dojo import PseudoDojoFamily, PseudoDojoConfiguration
 from aiida_pseudo.groups.family.sssp import SsspFamily, SsspConfiguration
+from aiida_pseudo.data.pseudo.upf import UpfData
```
sphuber, line 16: I'm trying to keep alphabetical order for imports

tests/cli/test_install.py
```diff
@@ -129,6 +130,26 @@ def test_install_family(run_cli_command, get_pseudo_archive):
...
+def test_install_family_folder(run_cli_command, filepath_pseudos):
+    """Test ``aiida-pseudo install family` from folder`."""
+    label = 'family_test'
+    description = 'description'
+    dirpath = filepath_pseudos()
+    options = ['-D', description, dirpath, label]
+
+    result = run_cli_command(cmd_install_family, options)
+    assert f'installed `{label}`' in result.output
+    assert PseudoPotentialFamily.objects.count() == 2
```
sphuber, line 145: why should this be 2? The database is cleared before this test isn't it?

tests/cli/test_install.py
```diff
@@ -139,14 +160,15 @@ def test_install_family_url(run_cli_command):
...
     result = run_cli_command(cmd_install_family, options)
     assert f'installed `{label}`' in result.output
-    assert SsspFamily.objects.count() == 1
+    assert PseudoPotentialFamily.objects.count() == 1
 
-    family = SsspFamily.objects.get(label=label)
-    assert family.__class__ is SsspFamily
+    family = PseudoPotentialFamily.objects.get(label=label)
+    assert isinstance(family.pseudos['Si'], UpfData)
+    #assert family.__class__ is SsspFamily
```
sphuber, line 171:
```suggestion
    assert family.__class__ is PseudoPotentialFamily
```

aiida_pseudo/cli/install.py
```diff
@@ -23,43 +24,56 @@ def cmd_install():
 
 
 @cmd_install.command('family')
-@click.argument('archive', type=types.FileOrUrl(mode='rb'))
+@click.argument('archive_or_folder', type=types.PathOrUrl(exists=True, file_okay=True))
 @click.argument('label', type=click.STRING)
 @options_core.DESCRIPTION(help='Description for the family.')
 @options.ARCHIVE_FORMAT()
-@options.FAMILY_TYPE()
+@options.PSEUDOS_TYPE()
```
sphuber, line 31:
```suggestion
@options.PSEUDO_TYPE()
```

aiida_pseudo/cli/install.py
```diff
@@ -23,43 +24,56 @@ def cmd_install():
...
-    # name in a different way.
-    try:
-        suffix = os.path.basename(archive.url)
-    except AttributeError:
-        suffix = os.path.basename(archive.name)
-
-    # Write the content of the archive to a temporary file, because `create_family_from_archive` does currently not
-    # accept filelike objects because the underlying `shutil.unpack_archive` does not. Likewise, `unpack_archive` will
-    # attempt to deduce the archive format from the filename extension, so it is important we maintain the original
-    # filename. Of course if this fails, users can specify the archive format explicitly wiht the corresponding option.
```
sphuber, line 56: Please reinstate this comment as it is not obvious why we need to preserve the suffix

aiida_pseudo/cli/install.py
```diff
@@ -23,43 +24,56 @@ def cmd_install():
...
-    with tempfile.NamedTemporaryFile(mode='w+b', suffix=suffix) as handle:
-        shutil.copyfileobj(archive, handle)
-        handle.flush()
+    # `archive_or_folder` is a simple string, containing the name of the folder / file / url location.
 
+    if os.path.isdir(archive_or_folder):
```
sphuber, line 51: We should start using `pathlib` exclusively for filepath manipulations.

aiida_pseudo/cli/install.py
```diff
@@ -23,43 +24,56 @@ def cmd_install():
...
+    elif os.path.isfile(archive_or_folder):
+        archive = archive_or_folder
         with attempt('unpacking archive and parsing pseudos... ', include_traceback=traceback):
-            family = create_family_from_archive(family_type, label, pathlib.Path(handle.name), fmt=archive_format)
+            family = create_family_from_archive(
+                PseudoPotentialFamily, label, pathlib.Path(archive), fmt=archive_format, pseudo_type=pseudos_type
+            )
+    else:
+        # The file of the url must be copied to a local temporary file. Maybe better ways to do it?
+        with urllib.request.urlopen(archive_or_folder) as archive:
```
sphuber, line 64: You are not checking if the opening of the URL actually works and so this will just except if there is anything wrong (I realize the original code also didn't check but it would be good to add now). Should probably use the `requests` library instead which makes it easy to check for problems. See the `download_sssp` for an example.

aiida_pseudo/cli/install.py
```diff
@@ -23,43 +24,56 @@ def cmd_install():
...
 @decorators.with_dbenv()
-def cmd_install_family(archive, label, description, archive_format, family_type, traceback):  # pylint: disable=too-many-arguments
-    """Install a standard pseudo potential family from an ARCHIVE on the local file system or from a URL.
+def cmd_install_family(archive_or_folder, label, description, archive_format, pseudos_type, traceback):  # pylint: disable=too-many-arguments
+    """Install a standard pseudo potential family from a FOLDER or an ARCHIVE (on the local file system or from a URL).
 
-    The command will attempt to infer the archive format from the filename extension of the ARCHIVE. If this fails, the
-    archive format can be specified explicitly with the archive format option, which will also display which formats
-    are supported.
+    The command will attempt first to recognize the passed `archive_or_folder` as a folder in the local system. If not,
```
sphuber, line 37:
```suggestion
    The command will attempt first to recognize the passed ARCHIVE_FOLDER as a folder in the local system. If not,
```
Since these docstrings will be rendered by `click` and are mostly intended for users of the CLI, I prefer to keep a reference to the params by their "metavar" which is formatted by `click`. The reason is that for users in the `--help` it will show as `ARCHIVE_OR_FOLDER` and so this doesn't correspond to `archive_folder`. For developers the inconsistency is less bad because they know that the function argument is turned into all-caps for the help string

sphuber (changes requested):
Thanks @bosonie . I have given it a first pass. I am wondering if we should reuse the `PathOrUrl` from `aiida-core`. It checks whether the remote URL can be fetched but than discards the result. Maybe it would be best to create our own custom type that does everything in one go as well as validation. But I can look at this myself later.

bosonie:
> > I'm also noticing now that
> > ```
> > aiida-pseudo install family https://legacy-archive.materialscloud.org/file/2018.0001/v4/SSSP_1.0_PBE_efficiency.tar.gz label
> > ```
> > 
> > 
> > that should be tested in `tests/cli/test_install.py` does actually fail in real life. Not because of a problem in the CLI command, but for failing of the underline `family_type.create_from_folder`. Is it expected?
> 
> How is it failing exactly? And is it failing on the `master` branch or with your patch? Or both?

```
Info: unpacking archive and parsing pseudos...  [FAILED]
Critical: `<class 'aiida_pseudo.data.pseudo.pseudo.PseudoPotentialData'>` constructor did not define the element and could not parse a valid element symbol from the filename `{filename}` either. It should have the format `ELEMENT.EXTENSION`
```
Also in master

sphuber:
> Also in master

I see, well the problem should actually be solved by your PR. Try to run the following on your branch:
```
aiida-pseudo install family https://legacy-archive.materialscloud.org/file/2018.0001/v4/SSSP_1.0_PBE_efficiency.tar.gz label -P pseudo.upf
```
As stated in the docstring, the base family type `PseudoPotentialFamily` will use the basic `PseudoPotentialData` for the pseudopotentials. The `PseudoPotentialData` is a base class for "any" format, but it still needs an element. For now, the way I coded it, it only tries to parse this from the filename, because the file content could be anything and there is no way how to parse it for the element. So we impose a strict naming scheme for the filename `{ELEMENT}.extension`. If this is not the case, we cannot parse the element and so this error is raised. Apparently the `SSSP_1.0_PBE_efficiency.tar.gz` archive contains a file that doesn't follow this convention and so the element cannot be parsed, hence the error. I am a bit hesitant to start making the parsing more clever because there will always be edge cases that we won't catch and this CLI command was supposed to be just a last way out. One can always use the API to directly to create a family however they want.

aiida_pseudo/cli/install.py
bosonie: This is already checked by `PathOrUrl` I believe.

aiida_pseudo/cli/params/types.py
bosonie: This would make fail the tests `tests/cli/test_list.py::test_list_filter`and `tests/cli/test_list.py::test_list_filter_no_result`

bosonie:
> I see, well the problem should actually be solved by your PR. Try to run the following on your branch:
Indeed!

aiida_pseudo/cli/params/types.py
sphuber: Ah I remember now, I added the original value that the user specifies on the param, because the converted value will be the class that correspond to the plugin. However, in the actual `list` command, I need the original entry point. Ok, good to keep then! :smile:

bosonie:
I applied all the changes a part form the one I commented.
I did not created our own custom so far. @sphuber let me know if you wish me to look into it.
Ideally I would love to merge this soon and have a release as well. I will use this features in `aiida-siesta` and I'm close to a release there as well.

sphuber:
> Ideally I would love to merge this soon and have a release as well. I will use this features in `aiida-siesta` and I'm close to a release there as well.

Just out of curiosity, since this PR just changes the CLI, why would you need this for `aiida-siesta`? And there are still a few unaddressed comments. I will comment on those specifically so you know which ones.

tests/cli/test_install.py
sphuber: I think there are still unnecessary empty lines here

tests/cli/test_install.py
sphuber: Still don't really understand why this should be 2? Maybe because this test is not properly decorated (because of the empty lines after the decorator with the fixture `clean_db`)

bosonie:
> Just out of curiosity, since this PR just changes the CLI, why would you need this for `aiida-siesta`? And there are still a few unaddressed comments. I will comment on those specifically so you know which ones.

Because in the package we deliver also examples and material for a tutorial. They require the creation of pseudo families and I do not want to make the topic more complicated than necessary.

tests/cli/test_install.py
bosonie: Sorry I forgot to check about this. Will do now. All the rest has been addressed.

bosonie: The decorator was misplaced. I'll push now the fix. Sorry!

sphuber:
> Because in the package we deliver also examples and material for a tutorial. They require the creation of pseudo families and I do not want to make the topic more complicated than necessary.

But you guys use psml right, and pseudos from Pseudo Dojo? So why can't you use `aiida-pseudo install pseudo-dojo`?

bosonie:
> But you guys use psml right, and pseudos from Pseudo Dojo? So why can't you use `aiida-pseudo install pseudo-dojo`?

No, `psml` is a new addition and will be the standard in the future. The current "stable" version of siesta still not support `psml`, only `psf`.

aiida_pseudo/cli/install.py
```diff
@@ -23,43 +24,60 @@ def cmd_install():
 
 
 @cmd_install.command('family')
-@click.argument('archive', type=types.FileOrUrl(mode='rb'))
+@click.argument('archive_or_folder', type=types.PathOrUrl(exists=True, file_okay=True))
 @click.argument('label', type=click.STRING)
 @options_core.DESCRIPTION(help='Description for the family.')
 @options.ARCHIVE_FORMAT()
-@options.FAMILY_TYPE()
+@options.PSEUDO_TYPE()
```
mbercx, line 29: I'm not sure if we should remove `FAMILY_TYPE` input here. I was trying to install a family as a `CutoffsFamily`, since the basic `PseudoPotentialFamily` class doesn't have the `RecommendedCutoffMixin` as a parent class, and I wanted to add recommended cutoffs to be used with the `get_builder_from_protocol()` method.

mbercx (commented):
Thanks @bosonie! I tried it out on a folder with pseudos I actually needed for a project, so you saved me the trouble of putting them in an archive! ;)

I just had one comment regarding removing the `FAMILY_TYPE` input. Although I do understand that if someone wants to install an `SsspFamily` or `PseudoDojoFamily`, they can use the corresponding command, I actually just now noticed that the base `PseudoPotentialFamily` doesn't have the `RecommendedCutoffMixin` as a parent class. So any pseudos installed with the command currently cannot use this functionality. There is the `CutoffsFamily` class, which @sphuber added mainly for testing purposes, but I'm thinking that perhaps we should make it possible for users to install this class type with the `install family` command. Or, is there any reason _not_ to give the base `PseudoPotentialFamily` class the methods for adding and using recommended cutoffs?

aiida_pseudo/cli/install.py
mbercx: Also related to #78.

sphuber:
@bosonie I have added a commit that implements our own version of `PathOrUrl` that makes the behavior of the command more consistent with respect to error handling. I have updated and simplified the implementation of the `install family` command accordingly and. If you are ok with the change, I can merge this. One more thing I realized while working on this, is that we probably should keep the option to change the family type. Creating an `SsspFamily` doesn't make sense, but I have added the `CutoffsFamily` and this will allow one to then add custom recommended cutoffs which may be useful. I will add this now

sphuber:
Haha, guess @mbercx agrees with me on that last part. I am not sure if we should add the `RecommendedCutoffMixin` to the `PseudoPotentialFamily` base class. Especially since we already have the `CutoffsFamily`, which indeed I mainly added for testing purposes, but now with the fully fledged `install family` command, it starts to make sense for real use. Maybe we should rename it slightly though to `CutoffsPseudoPotentialFamily` to still make it plainly obvious that it is a pseudopotential family. I would also still check the type and only allow `pseudo.family` and `pseudo.family.cutoffs`. I don't think we want to allow creating pseudo-dojo or SSSP families through this command. Those respective commands already do that and if there are problems with internet, they even have the option to install directly from archive on disk.

sphuber:
@bosonie and @mbercx added the implementation for the family type. Please have a look

mbercx:
> Haha, guess @mbercx agrees with me on that last part. 

😁 

> I am not sure if we should add the `RecommendedCutoffMixin` to the `PseudoPotentialFamily` base class. 

Sure, I do appreciate the cleanliness of a mixin class. I'm just wondering if there is a use case for a pseudo family _without_ recommended cutoffs. There probably is, I'm just too in ❤️  with the `get_builder_from_protocol` method.

> Especially since we already have the `CutoffsFamily`, which indeed I mainly added for testing purposes, but now with the fully fledged `install family` command, it starts to make sense for real use. Maybe we should rename it slightly though to `CutoffsPseudoPotentialFamily` to still make it plainly obvious that it is a pseudopotential family. 

Fully agree 👍 

> I would also still check the type and only allow `pseudo.family` and `pseudo.family.cutoffs`. I don't think we want to allow creating pseudo-dojo or SSSP families through this command. Those respective commands already do that and if there are problems with internet, they even have the option to install directly from archive on disk.

Although I do agree, I did have a use case for installing a different version of the SSSP today, since there is a small issue with the header of the lanthanide SSSP files and I wanted to test the fixed versions. To circumvent the restriction on the label, I installed the fixed version as `SSSP/1.0/efficiency` 😬 . Anyways, this is a rare developer use case, and can be easily accomplished by installing a `CutoffsPseudoPotentialFamily` in the future.

mbercx (approved):
Looks good, thanks @sphuber! Just a small comment that we should perhaps help the user to the right command if they want to install `sssp` or `pseudo_dojo` families.

Also: Maybe we should immediately adapt the docstring of `CutoffsFamily` now that it is a proper class with use cases beyond testing? :)

aiida_pseudo/cli/params/types.py
```diff
@@ -78,6 +86,9 @@ def convert(self, value, _, __):
         except exceptions.EntryPointError as exception:
             raise click.BadParameter(f'`{value}` is not an existing group plugin.') from exception
 
+        if self.exclude and value in self.exclude:
+            raise click.BadParameter(f'`{value}` is not an accepted value for this option.')
```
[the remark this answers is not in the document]
sphuber: Just would be tricky to do that here. Sure now the type is used only in one command and there I exclude two values that each have their own dedicated command, but that might not always apply. So putting that logic here feels too much like hard-coding. I think we should rather just add this to the README/documentation.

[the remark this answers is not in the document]
mbercx: Yeah, I see your point. I guess I better get to work on that documentation! 😅

sphuber:
OK, we are starting to cram a bit much into a single PR. I am going to factor out the recent additions and then rebase this so it includes mostly the original idea of fixing `install family`.

bosonie:
All good from my side! Unlike @mbercx, I was always familiar with families without recommended cutoffs. The good side of a collaborative effort!
Thanks a million!

aiida_pseudo/cli/install.py
```diff
@@ -23,46 +22,70 @@ def cmd_install():
...
+    depending on the chosen pseudopotential type (see the option `-P/--pseudo-type`) there can be additional
+    requirements on the pseudopotential file and filename format.
+
+    If the ARCHIVE corresponds to a (compressed) archive, the command will attempt to infer the archive format from the
+    filename extension of the ARCHIVE. If this fails, the archive format can be specified explicitly with the archive
+    format option `-F/--archive-format`, which will also display which formats are supported. These format suffixes
+    follow the naming of the `shutil.unpack_archive` standard library method.
+
+    Once the ARCHIVE is downloaded, uncompressed and unarchived into a directory on the local file system, the command
+    will create a `PseudoPotentialFamily` instance where the type of the pseudopotential data nodes that are stored
```
mbercx, line 50: Now that the `FAMILY_TYPE` option has been reinstated, we should probably also revert back to explaining that the pseudo family instance class can be set.

aiida_pseudo/cli/install.py
```diff
@@ -23,46 +22,70 @@ def cmd_install():
...
+    within it is set through the `-P/--pseudo-type` option. If the default `pseudo` (which corresponds to the data
+    plugin `PseudoPotentialData`) the pseudopotential files in the archive *have* to have filenames that strictly follow
```
mbercx, line 52:
```suggestion
    plugin `PseudoPotentialData`) is used, the pseudopotential files in the archive *have* to have filenames that strictly follow
```

aiida_pseudo/cli/install.py
```diff
@@ -23,46 +22,70 @@ def cmd_install():
...
+    elif isinstance(archive, pathlib.Path) and archive.is_file():
         with attempt('unpacking archive and parsing pseudos... ', include_traceback=traceback):
-            family = create_family_from_archive(family_type, label, pathlib.Path(handle.name), fmt=archive_format)
+            family = create_family_from_archive(
+                family_type, label, archive, fmt=archive_format, pseudo_type=pseudo_type
+            )
+    else:
+        # At this point, we can assume that it is not a valid filepath on disk, but rather a URL and the ``archive``
+        # variable will contain the result objects from the ``requests`` library. The validation of the URL will already
+        # have been done by the ``PathOrUrl`` parameter type, so the URL is reachabel. The content of the URL must be
```
mbercx, line 70:
```suggestion
        # have been done by the ``PathOrUrl`` parameter type, so the URL is reachable. The content of the URL must be
```

aiida_pseudo/cli/install.py
```diff
@@ -23,46 +22,70 @@ def cmd_install():
...
+            handle.flush()
+
+            with attempt('unpacking archive and parsing pseudos... ', include_traceback=traceback):
+                family = create_family_from_archive(
+                    family_type, label, pathlib.Path(handle.name), fmt=archive_format, pseudo_type=pseudo_type
+                )
 
     family.description = description
-    echo.echo_success(f'installed `{label}` containing {family.count()} pseudo potentials')
+    echo.echo_success(f'installed `{label}` containing {family.count()} pseudopotentials')
```
mbercx, line 88: image

mbercx (approved):
Thanks @sphuber! Just gave the PR a final pass, and have left some minor comments. Merge at your discretion! 

> Unlike @mbercx, I was always familiar with families without recommended cutoffs. 

I guess I'm a little spoiled. 😅

aiida_pseudo/cli/install.py
sphuber: I learn from the best

sphuber (approved):

sphuber:
Thanks a lot @bosonie !

PR #86, opened 2021-05-06
`RecommendedCutoffMixin`: improve error messages

### LINKED ISSUE #84: `RecommendedCutoffMixin`: improve warning in case final stringency is deleted
Currently, when you remove the final cutoffs stringency from a pseudo family, the warning is rather nonsensical:

```Python
/home/mbercx/envs/aiida-sirius/code/aiida-pseudo/aiida_pseudo/groups/mixins/cutoffs.py:242: UserWarning: `normal` was the default stringency of this family. Please set one of () as the new default stringency with the `set_default_stringency` method.
  warnings.warn(warning)
```

In case the stringency that is removed is the final one, this message should probably be something like:

```Python
/home/mbercx/envs/aiida-sirius/code/aiida-pseudo/aiida_pseudo/groups/mixins/cutoffs.py:242: UserWarning: `normal` was the default stringency of this family. Since no other stringencies are defined for this family, no new default can be specified.
  warnings.warn(warning)
```

### LINKED ISSUE #85: `RecommendedCutoffMixin`: improve error message when element is not in family
When trying to use the `get_recommended_cutoffs` method to obtain cutoffs for an element that isn't in the pseudo family, the error message is not very clear:

```Python
KeyError                                  Traceback (most recent call last)
<ipython-input-20-f8b0a3e88790> in <module>
----> 1 pseudo_family.get_recommended_cutoffs(elements='Nd')

~/envs/aiida-sirius/code/aiida-pseudo/aiida_pseudo/groups/mixins/cutoffs.py in get_recommended_cutoffs(self, elements, structure, stringency, unit)
    304                 values = {k: U.Quantity(v, current_unit).to(unit).to_tuple()[0] for k, v in cutoffs[element].items()}
    305             else:
--> 306                 values = cutoffs[element]
    307 
    308             cutoffs_wfc.append(values['cutoff_wfc'])

KeyError: 'Nd'
```

It would be better to catch this `KeyError` and return `ValueError` with a more elucidating message.

mbercx:
Fixes #84 
Fixes #85 

Improve the error message in two cases:

* When deleting the final stringency with the `delete_cutoffs` method.
* An element requested with the `get_recommended_cutoffs` method does
not have a pseudo in the pseudo family.

### DIFF
--- a/aiida_pseudo/groups/mixins/cutoffs.py
+++ b/aiida_pseudo/groups/mixins/cutoffs.py
@@ -280,6 +280,7 @@ class RecommendedCutoffMixin:
         :return: tuple of recommended wavefunction and density cutoff.
         :raises ValueError: if the requested stringency is not defined for this family.
         :raises ValueError: if optional unit specified is invalid.
+        :raises ValueError: if the family does not have a pseudo for one of the elements (of the structure).
         """
         if (elements is None and structure is None) or (elements is not None and structure is not None):
             raise ValueError('at least one and only one of `elements` or `structure` should be defined')
--- a/aiida_pseudo/groups/mixins/cutoffs.py
+++ b/aiida_pseudo/groups/mixins/cutoffs.py
@@ -220,7 +220,7 @@ class RecommendedCutoffMixin:
         try:
             if stringency == self.get_default_stringency():
                 self.delete_extra('_default_stringency')
-                warning += f'`{stringency}` was the default stringency of this family. '
+                warning += f'`{stringency}` was the default stringency of this family.'
                 assign_new_default = True
             else:
                 assign_new_default = False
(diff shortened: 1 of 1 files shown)


aiida_pseudo/groups/mixins/cutoffs.py
```diff
@@ -299,6 +303,9 @@ def get_recommended_cutoffs(self, *, elements=None, structure=None, stringency=N
 
         for element in symbols:
 
+            if element not in cutoffs:
+                raise ValueError(f'family does not contain a pseudo for element `{element}`.')
```
sphuber, line 307: can you add this to the docstring as a potential raise please

sphuber (commented):
Thanks @mbercx , would just add the new exception to docstring but that is a minor thing

aiida_pseudo/groups/mixins/cutoffs.py
mbercx: Of course, silly me!

sphuber (approved):
Nais!

Issue #87, opened 2021-05-06
`PseudoPotentialData`: improve error message when user try to instanciate with an abs path 


bosonie:
Users of families like `UpfData` `PsmlData` had in the past (before the introduction of aiida-pseudo package) the possibility to pass either a filelike object or the absolute path of the file to instantiate these classes. The absolute path is not supported any-more.
If a user now tries to pass the absolute path, the printed error is:
```
 AttributeError: 'str' object has no attribute 'seek'
```
which is not very informative.
I would suggest to catch absolute paths and print a more informative error.

Also note that  `PseudoPotentialData` inherit from `SinglefileData` that do accepts both  filelike object and absolute paths. So even more expert users might be confused by the error.

PR #90, opened 2021-05-07
Improve the error message if a file path is passed to get_or_create

zooks97:
In response to #87. I'm not entirely sure that this is where the error is most likely, as it seems like `set_file` may also be the culprit of @bosonie 's error message.


sphuber:
Thanks @zooks97 . There are more places where this can be raised though. And I think here it would be better to go against the Python mantra of asking for forgiveness rather than permission and actually explicitly check the type of the argument. In this case the `AttributeError` could be thrown for another reason and we might give the wrong error message. I was going to submit a PR for this but I was still considering the request of also supporting a filepath instead of a stream. I will get back to this soon.

PR #95, opened 2021-05-08
Update README.md

mbercx:
The README still had a lot of information that has now been moved to the
documentation. Here we clean up the document and simply add another
link to the documentation, along with the license and acknowledgements.


sphuber:
How about we put the acknowledgments in table: https://github.com/aiidateam/aiida-pseudo/tree/fix/clean-README
This looks a lot cleaner since the images are inline

mbercx:
Table looks great, @sphuber! I also tried making the table background white, since now the logos are hard to read in dark mode, but apparently it's not that easy. It's seems very tricky to make logos look good for both white and dark mode atm:

https://github.community/t/support-theme-context-for-images-in-light-vs-dark-mode/147981

So I simply undid the change. 🤷

sphuber (approved):

PR #97, opened 2021-05-09
CLI: Add support for units to `family show`

### LINKED ISSUE #96: CLI: Add support for units to `family show`
We've added support for units in https://github.com/aiidateam/aiida-pseudo/pull/57, but forgot to update the `family show` command. This means that the user still sees the hardcoded default (`eV`), even when the unit of the configured recommended cutoffs is different:

```console
$ aiida-pseudo family show SSSP/1.1/PBE/efficiency 
Element    Pseudo                          MD5                                 Wavefunction (eV)    Charge density (eV)
---------  ------------------------------  --------------------------------  -------------------  ---------------------
Ag         Ag_ONCV_PBE-1.0.oncvpsp.upf     94f47bd0669c641108e45594df92fabc                 50.0                  200.0
Al         Al.pbe-n-kjpaw_psl.1.0.0.UPF    cfc449ca30b5f3223ec38ddd88ac046d                 30.0                  240.0
Ar         Ar_ONCV_PBE-1.1.oncvpsp.upf     46d28409cdd246843f76b7675277a949                 60.0                  240.0
As         As.pbe-n-rrkjus_psl.0.2.UPF     767315de957beeeb34f87d97bf945c8f                 35.0                  280.0
```

We should:

* Make sure the unit shown in the header is correct.
* Add the `--unit` option to the `family show` command.

mbercx:
Fixes #96 

Although we now support specifying units for the recommended cutoffs,
the `family show` method had not been adapted for this feature.

Here we add the `-u/--unit` option to the `family show` method, making
sure to show the correct unit in the table header as well as adapting
the values in the columns by supplying the unit to the
`get_recommended_cutoffs` method.

There is also some refactoring regarding the validation of the
stringencies. The `RecommendedCutoffMixin.validate_stringency` method
was not being used, so we adapt the code for several methods with the
stringency input to rely on this validation method. This way we have a
consistent and succinct error message in case the user requests a
stringency that has not been configured.

### DIFF
--- a/aiida_pseudo/cli/family.py
+++ b/aiida_pseudo/cli/family.py
@@ -20,21 +20,26 @@ def cmd_family():
 @cmd_family.command('show')
 @arguments.PSEUDO_POTENTIAL_FAMILY()
 @options.STRINGENCY()
+@options.UNIT(default=None)
 @options_core.RAW()
 @decorators.with_dbenv()
-def cmd_family_show(family, stringency, raw):
+def cmd_family_show(family, stringency, unit, raw):
     """Show details of pseudo potential family."""
     from tabulate import tabulate
 
     if isinstance(family, RecommendedCutoffMixin):
 
-        if stringency is not None and stringency not in family.get_cutoff_stringencies():
-            raise click.BadParameter(f'`{stringency}` is not defined for family `{family}`.', param_hint='stringency')
+        try:
+            family.validate_stringency(stringency)
+        except ValueError as exception:
+            raise click.BadParameter(f'{exception}', param_hint="'-s' / '--stringency'")
 
-        headers = ['Element', 'Pseudo', 'MD5', 'Wavefunction (eV)', 'Charge density (eV)']
+        unit = unit or family.get_cutoffs_unit(stringency)
+
+        headers = ['Element', 'Pseudo', 'MD5', f'Wavefunction ({unit})', f'Charge density ({unit})']
         rows = [[
             pseudo.element, pseudo.filename, pseudo.md5,
-            *family.get_recommended_cutoffs(elements=pseudo.element, stringency=stringency)
+            *family.get_recommended_cutoffs(elements=pseudo.element, stringency=stringency, unit=unit)
         ] for pseudo in family.nodes]
     else:
         headers = ['Element', 'Pseudo', 'MD5']
@@ -75,11 +80,6 @@ def cmd_family_cutoffs_set(family, cutoffs, stringency, unit):  # noqa: D301
     """
     if not isinstance(family, RecommendedCutoffMixin):
         raise click.BadParameter(f'family `{family}` does not support recommended cutoffs to be set.')
-
-    try:
-        family.validate_cutoffs_unit(unit)
-    except ValueError as exception:
-        raise click.BadParameter(f'{exception}', param_hint='UNIT')
 
     try:
         data = json.load(cutoffs)
--- a/aiida_pseudo/groups/mixins/cutoffs.py
+++ b/aiida_pseudo/groups/mixins/cutoffs.py
@@ -2,6 +2,8 @@
 """Mixin that adds support of recommended cutoffs to a ``Group`` subclass, using its extras."""
 import warnings
 
+from typing import Optional
+
 from aiida.common.lang import type_check
 from aiida.plugins import DataFactory
 
@@ -81,17 +83,23 @@ class RecommendedCutoffMixin:
         if not U.Quantity(1, unit).check('[energy]'):
             raise ValueError(f'`{unit}` is not a valid energy unit.')
 
-    def validate_stringency(self, stringency: str) -> None:
+    def validate_stringency(self, stringency: Optional[str]) -> None:
         """Validate a cutoff stringency.
 
+        Check if the stringency is defined for the family. If no stringency input is passed, the method checks if a
+        default stringency has been set.
+
         :param stringency: the cutoff stringency to validate.
-        :raises ValueError: if stringency is None or the family does not define cutoffs for the specified stringency.
+        :raises ValueError: if default stringency has not been defined.
+        :raises ValueError: if the family does not define cutoffs for the specified stringency.
         """
         if stringency is None:
-            raise ValueError('defining a stringency is required.')
-
-        if stringency not in self.get_cutoff_stringencies():
-            raise ValueError(f'stringency `{stringency}` is not defined for this family.')
+            self.get_default_stringency()
+        elif stringency not in self.get_cutoff_stringencies():
+            raise ValueError(
+                f'stringency `{stringency}` is not one of the available cutoff stringencies for this family: '
+                f'{self.get_cutoff_stringencies()}.'
+            )
 
     def _get_cutoffs_dict(self) -> dict:
         """Return the cutoffs dictionary that maps the stringencies to the recommended cutoffs.
@@ -125,12 +133,7 @@ class RecommendedCutoffMixin:
         :raises ValueError: if the provided default stringency is not in the tuple of available cutoff stringencies for
             the pseudo family.
         """
-        if default_stringency not in self.get_cutoff_stringencies():
-            raise ValueError(
-                'provided default stringency not in tuple of available cutoff stringencies: '
-                f'{self.get_cutoff_stringencies()}.'
-            )
-
+        self.validate_stringency(default_stringency)
         self.set_extra(self._key_default_stringency, default_stringency)
 
     def get_cutoff_stringencies(self) -> tuple:
@@ -182,7 +185,7 @@ class RecommendedCutoffMixin:
         if len(cutoffs_dict) == 1:
             self.set_default_stringency(stringency)
 
-    def get_cutoffs(self, stringency=None) -> dict:
+    def get_cutoffs(self, stringency: str = None) -> dict:
         """Return a set of cutoffs for the given stringency.
 
         :param stringency: optional stringency for which to retrieve the cutoffs. If not specified, the default
@@ -190,14 +193,11 @@ class RecommendedCutoffMixin:
         :raises ValueError: if no stringency is specified and no default stringency is defined for the family.
         :raises ValueError: if the requested stringency is not defined for this family.
         """
+        self.validate_stringency(stringency)
         stringency = stringency or self.get_default_stringency()
+        return self._get_cutoffs_dict()[stringency]
 
-        try:
-            return self._get_cutoffs_dict()[stringency]
-        except KeyError as exception:
-            raise ValueError(f'stringency `{stringency}` is not defined for this family.') from exception
-
-    def delete_cutoffs(self, stringency) -> None:
+    def delete_cutoffs(self, stringency: str) -> None:
         """Delete the recommended cutoffs for a specified stringency.
 
         .. note: If, after the cutoffs have been deleted, there is only one stringency defined for the pseudo family,
@@ -207,8 +207,7 @@ class RecommendedCutoffMixin:
         :param stringency: stringency for which to delete the cutoffs.
         :raises ValueError: if the requested stringency is not defined for this family.
         """
-        if stringency not in self.get_cutoff_stringencies():
-            raise ValueError(f'stringency `{stringency}` is not defined for this family.')
+        self.validate_stringency(stringency)
 
         cutoffs_dict = self._get_cutoffs_dict()
         cutoffs_dict.pop(stringency)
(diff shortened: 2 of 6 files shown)


aiida_pseudo/groups/mixins/cutoffs.py
```diff
@@ -85,13 +85,13 @@ def validate_stringency(self, stringency: str) -> None:
         """Validate a cutoff stringency.
 
         :param stringency: the cutoff stringency to validate.
-        :raises ValueError: if stringency is None or the family does not define cutoffs for the specified stringency.
+        :raises ValueError: if the family does not define cutoffs for the specified stringency.
         """
-        if stringency is None:
```
mbercx, line 90: I'm not sure why this had to raise, since `stringency` is not an optional input, and type hinting indicates it has to be a string.

aiida_pseudo/cli/family.py
```diff
@@ -20,21 +20,28 @@ def cmd_family():
...
-            raise click.BadParameter(f'`{stringency}` is not defined for family `{family}`.', param_hint='stringency')
+        if stringency is not None:
+            try:
+                family.validate_stringency(stringency)
+            except ValueError as exception:
+                raise click.BadParameter(f'{exception}', param_hint='stringency')
 
-        headers = ['Element', 'Pseudo', 'MD5', 'Wavefunction (eV)', 'Charge density (eV)']
+        if unit is None:
+            unit = family.get_cutoffs_unit(stringency)
```
sphuber, line 39: This should now fail if no stringency is defined and there is no default defined. We could simply try/catch this `ValueError`. I wonder if there is a better way because this concept will come back more often where we have to separately try/catch the `validate_stringency` and some other function, like `get_cutoffs_unit`. Maybe we should at the very least allow `None` for `stringency` in `validate_stringency` and then if it is `None` _it_ will already try to get the default and raise if it doesn't exist.

sphuber, line 39: If we go with my other suggestion to validate the default in `validate_stringency`, we can turn this into
```suggestion
        unit = unit or family.get_cutoffs_unit(stringency)
```

aiida_pseudo/groups/mixins/cutoffs.py
```diff
@@ -85,13 +85,13 @@ def validate_stringency(self, stringency: str) -> None:
...
-            raise ValueError('defining a stringency is required.')
-
         if stringency not in self.get_cutoff_stringencies():
-            raise ValueError(f'stringency `{stringency}` is not defined for this family.')
+            raise ValueError(
+                f'stringency \'{stringency}\' is not one of the available cutoff stringencies for this family: '
```
sphuber, line 92:
```suggestion
                f'stringency `{stringency}` is not one of the available cutoff stringencies for this family: '
```

aiida_pseudo/groups/mixins/cutoffs.py
```diff
@@ -190,12 +185,11 @@ def get_cutoffs(self, stringency=None) -> dict:
...
-        stringency = stringency or self.get_default_stringency()
-
-        try:
-            return self._get_cutoffs_dict()[stringency]
-        except KeyError as exception:
-            raise ValueError(f'stringency `{stringency}` is not defined for this family.') from exception
+        if stringency is None:
+            stringency = self.get_default_stringency()
+        else:
+            self.validate_stringency(stringency)
```
sphuber, line 191: I would say that code like this is another argument in favor of `validate_stringency` accepting `None` and just calling `get_default_stringency` in that case.

aiida_pseudo/groups/mixins/cutoffs.py
```diff
@@ -257,19 +250,20 @@ def get_cutoffs_unit(self, stringency: str = None) -> str:
...
 
         try:
             return self._get_cutoffs_unit_dict()[stringency]
-        except KeyError as exception:
+        except KeyError:
             # Workaround to deal with pseudo families installed in v0.5.0 - Set default unit in case it is not in extras
-            if stringency in self.get_cutoff_stringencies():
```
sphuber, line 266: do we no longer need this conditional?

tests/cli/test_family.py
```diff
@@ -131,3 +131,33 @@ def test_family_show_raw(clear_db, run_cli_command, get_pseudo_family):
...
+        for unit in ['eV', 'hartree', 'aJ']:
+            result = run_cli_command(cmd_family_show, [family.label, option, unit])
+            cutoffs = family.get_recommended_cutoffs(elements='Ar', unit=unit)
+            header_fields = result.output_lines[0].split()
+            assert header_fields[4] == f'({unit})'
+            assert header_fields[4] == f'({unit})'
+
+            values_fields = result.output_lines[2].split()
+            assert round(cutoffs[0], 1) == float(values_fields[3])
+            assert round(cutoffs[1], 1) == float(values_fields[4])
```
sphuber, line 163: Would be good to parametrize this test over the various units. Looks like we have more unit tests ;)
Noticed afterwards that before the loop over units you also do a check for the default and there you check all lines for all elements, and for the unit specific test you only check the `Ar` cutoffs. Maybe we can split these tests in two

```suggestion
@pytest.mark.usefixtures(('unit',), (None, 'eV', 'hartree', 'aJ'))
def test_family_show_unit(clear_db, run_cli_command, get_pseudo_family, unit):
    """Test the `-u/--unit` option."""
    elements = ['Ar', 'Kr']
    cutoff_dict = {'normal': {'Ar': {'cutoff_wfc': 50, 'cutoff_rho': 200}, 'Kr': {'cutoff_wfc': 25, 'cutoff_rho': 100}}}

    family = get_pseudo_family(cls=CutoffsPseudoPotentialFamily, elements=elements, cutoffs_dict=cutoff_dict, unit='Ry')

    # Test the default unit (Ry)
    result = run_cli_command(cmd_family_show, [family.label, '--raw'])

    for index, element in enumerate(elements):
        cutoffs = family.get_recommended_cutoffs(elements=element)
        fields = result.output_lines[index].split()
        assert_almost_equal(cutoffs[0], float(fields[3]))
        assert_almost_equal(cutoffs[1], float(fields[4]))

    # Test both option strings and several units
    for option in ['-u', '--unit']:
        options = [family.label, '--raw']
        
        if unit is not None:
            options.extend([option, unit])
            
        result = run_cli_command(cmd_family_show, options)
        cutoffs = family.get_recommended_cutoffs(elements='Ar', unit=unit)
        header_fields = result.output_lines[0].split()
        assert header_fields[4] == f'({unit})'
        assert header_fields[4] == f'({unit})'

        values_fields = result.output_lines[2].split()
        assert round(cutoffs[0], 1) == float(values_fields[3])
        assert round(cutoffs[1], 1) == float(values_fields[4])
```

sphuber (changes requested):
Thanks @mbercx , some suggestions

aiida_pseudo/groups/mixins/cutoffs.py
mbercx: I know we've been using backticks for this purpose a lot, but is there any reason to prefer backticks here over single quotes, since the stringency is actually a string?

aiida_pseudo/groups/mixins/cutoffs.py
mbercx: Yes, I agree! Muy limpio!

aiida_pseudo/groups/mixins/cutoffs.py
sphuber: Just for consistency. I don't think we need the literal single quotes to make it clear this is a string value and just having backticks everywhere to reflect values that were "dynamic" and should interest the caller specifically, seems useful to me. Also, backticks don't have to be escaped :grimacing:

aiida_pseudo/groups/mixins/cutoffs.py
sphuber: Well, because the following is allowed in Python:
```
In [1]: def test(a: str):
   ...:     print(a)
   ...: 

In [2]: test(None)
None

```
it is type _hinting_ and so not enforced. `None` is a valid option and if we don't accept it, we have to check it explicitly. 
The following:
```
In [4]: test()
TypeError                                 Traceback (most recent call last)
<ipython-input-4-fbd55f77ab7c> in <module>
----> 1 test()

TypeError: test() missing 1 required positional argument: 'a'
```
raises though.
This again comes back to our discussion recently on the protocols and the fact that `None` for an argument and the absence of an argument are not the same thing.

mbercx: Sure, I understand that someone could pass `None` and it would raise with a slightly more clear error message if we catch this, but this could also be said for the `unit` in `validate_cutoffs_unit`, or basically almost any input of any method. Someone could also pass a `tuple`, `list` etc. I just didn't understand why we had to catch this user error specifically, since it would still raise because `None` wouldn't be in the list of defined stringencies.

However, now that inputting `None` checks if there is a _default_ stringency set, the point is moot. 😄

aiida_pseudo/groups/mixins/cutoffs.py
mbercx: True, it's cleaner and plenty clear. Just decided to try it and see what happens. 😁

aiida_pseudo/groups/mixins/cutoffs.py
mbercx: Isn't this already checked now by `validate_stringency`?

aiida_pseudo/cli/family.py
mbercx: Indeed, good point and implemented.

Now we still will have quite a bit code duplication like this in the CLI:

```
    try:
        family.validate_stringency(stringency)
    except ValueError as exception:
        raise click.BadParameter(f'{exception}', param_hint='STRINGENCY')

    try:
        family.validate_cutoffs_unit(unit)
    except ValueError as exception:
        raise click.BadParameter(f'{exception}', param_hint='UNIT')
```

Can you pass a click argument to the initialisation of the option descriptor on the same command? Then we could pass `family` and do this validation in the option definition.

mbercx: I guess you can do this from the context. Don't spoil it for me, I'll figure it out. 😬

mbercx: Hmm, I though I could define a `ParamType` and use the `ctx` in e.g. the `convert` method to obtain the `family`. But the `args` and `params` attributes of the `ctx` seem to be empty, so I'm probably going to need to understand a bit more about the inner workings of `click` to figure this out.

tests/cli/test_family.py
mbercx: Oh, I only after executing your suggestion saw that you actually had made a GitHub suggestion. Oops! 😅

aiida_pseudo/cli/family.py
mbercx: In the end I only created the `UnitParamType` that has a convert method to check if the provided unit corresponds to the right quantity (energy in this case). The code here is adapted from the `validate_cutoffs_unit` method of the `RecommendedCutoffMixin` and make more general to also check for other quantities. Note that we could also use a decorator from `pint` for this purpose:

https://pint.readthedocs.io/en/stable/wrapping.html#checking-dimensionality

But perhaps our error messages are more succinct and clear as they are.

For the stringency, there is only 1 location where it needs to be validated in the CLI atm, since for the `family cutoffs set` command the stringency should _not_ be validated to be defined on the class already, of course. 🙄

mbercx: That said, if you know how to use the `family` argument in the `covert` method of the type of another option defined on a command, I would love to know. 😁

mbercx:
@sphuber this is ready for another round of review! I've still left some comments unresolved since there might still be some discussion.

aiida_pseudo/cli/family.py
sphuber: Yeah, this is a known limitation of the design of `click`. Essentially, it is not possible to have validators for parameters that depend on other parameters because the order in which parameters are processed is not well defined. I know that I went down this rabbit-hole a long time ago and came up empty-handed

aiida_pseudo/groups/mixins/cutoffs.py
sphuber: Ah, I see now how this used to work. All good

aiida_pseudo/cli/family.py
mbercx: I see. I'll give it another go if we start having a lot of duplicated code because of validation that requires other parameters. 👍

aiida_pseudo/groups/mixins/cutoffs.py
```diff
@@ -125,12 +133,7 @@ def set_default_stringency(self, default_stringency: str) -> None:
         :raises ValueError: if the provided default stringency is not in the tuple of available cutoff stringencies for
             the pseudo family.
         """
-        if default_stringency not in self.get_cutoff_stringencies():
-            raise ValueError(
-                'provided default stringency not in tuple of available cutoff stringencies: '
-                f'{self.get_cutoff_stringencies()}.'
-            )
-
+        self.validate_stringency(default_stringency)
```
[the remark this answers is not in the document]
mbercx: Yeah, I don't see a use case for unsetting the default stringency. Will adjust! 👍

sphuber (approved):

PR #102, opened 2021-06-07
Change filter to actually filter

QwertycowMoo:
The current filter doesn't actually filter, it requires this operator to filter correctly.


sphuber:
Thanks for the PR @QwertycowMoo . The `==` operator is implicit if none is specified, so the change you applied is identical to the original code in function. Do you have an example of usage where you think the code doesn't function properly? Because this code is also tested and the tests pass without issues.

QwertycowMoo:
Could not find a way to reproduce and forgot how to replicate initial issue. Closed for now and sorry for the inconvenience!

PR #104, opened 2021-08-11
👌 IMPROVE: links to materials cloud SSSP archive

### LINKED ISSUE #13: Update materials cloud archive URL's
When trying to install the default SSSP family, I ran into connectivity issues:

```
$ aiida-pseudo install sssp
Info: downloading selected pseudo potentials archive...  [FAILED]
Critical: HTTPSConnectionPool(host='legacy-archive.materialscloud.org', port=443): Max retries exceeded with url: /file/2018.0001/v4/SSSP_1.1_PBE_efficiency.tar.gz (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7fedc762c5f8>: Failed to establish a new connection: [Errno 110] Connection timed out',))
```

After touching base with @borellim, it turns out that the legacy archive had been shut down, since they didn't think anyone was still using it. It's back online now, but we should probably update the URL's so we are no longer relying on the legacy archive.

mbercx:
Fixes #13 

The current links used to download the SSSP pseudopotentials still point
to the legacy materials cloud archive, which are hosted on a server that
they want to shut down. Next to this, the archive now also contains new
patch versions for minor version 1.1, with 1.1.2 being the latest. Since
these patch versions usually implement bug fixes, there is no reason to
install an outdated patch version. So instead of allowing the user to
install specific patch versions, we'll simply stick to only using minor
versions and make sure that the latest patch version is installed.

Here we update the links to the new servers. However, we also add some
extra functionality to retrieve a mapping of the latest patch version
corresponding to the minor version of the configuration. Based on this
mapping, the correct url is constructed for the archive and metadata.

### DIFF
--- a/aiida_pseudo/cli/install.py
+++ b/aiida_pseudo/cli/install.py
@@ -111,19 +111,20 @@ def download_sssp(
 
     url_template = 'https://archive.materialscloud.org/record/file?filename={filename}&parent_id=19'
 
+    # Download the dictionary mapping of the minor versions to the latest corresponding patch versions. Since patch
+    # releases of the SSSP only contain bug fixes, there is no reason to have the user install an outdated patch
+    # version. So, the latest patch version of the minor version that is specified by the user is always installed.
     with attempt('downloading patch versions information... ', include_traceback=traceback):
         response = requests.get(url_template.format(filename='versions.yaml'))
         response.raise_for_status()
+        # The `version_mapping` is a dictionary that maps each minor version (key) to the latest patch version (value)
         version_mapping = yaml.load(response.content, Loader=yaml.SafeLoader)
+        patch_version = version_mapping[configuration.version]
 
-    echo.echo_info(f'Latest patch version found: {version_mapping[configuration.version]}')
+    echo.echo_info(f'Latest patch version found: {patch_version}')
 
-    archive_filename = SsspFamily.format_configuration_filename(
-        configuration, 'tar.gz', version_mapping[configuration.version]
-    )
-    metadata_filename = SsspFamily.format_configuration_filename(
-        configuration, 'json', version_mapping[configuration.version]
-    )
+    archive_filename = SsspFamily.format_configuration_filename(configuration, 'tar.gz', patch_version)
+    metadata_filename = SsspFamily.format_configuration_filename(configuration, 'json', patch_version)
 
     url_archive = url_template.format(filename=archive_filename)
     url_metadata = url_template.format(filename=metadata_filename)
--- a/aiida_pseudo/cli/install.py
+++ b/aiida_pseudo/cli/install.py
@@ -105,6 +105,7 @@ def download_sssp(
     :param filepath_archive: absolute filepath to write the pseudopotential archive to.
     :param filepath_metadata: absolute filepath to write the metadata file to.
     :param traceback: boolean, if true, print the traceback when an exception occurs.
+    :return: Latest patch version of the requested minor version
     """
     from aiida_pseudo.groups.family import SsspFamily
     from .utils import attempt
@@ -142,6 +143,8 @@ def download_sssp(
         with open(filepath_metadata, 'wb') as handle:
             handle.write(response.content)
             handle.flush()
+
+    return patch_version
 
 
 def download_pseudo_dojo(
@@ -202,7 +205,6 @@ def cmd_install_sssp(version, functional, protocol, download_only, traceback):
 
     configuration = SsspConfiguration(version, functional, protocol)
     label = SsspFamily.format_configuration_label(configuration)
-    description = f'SSSP v{version} {functional} {protocol} installed with aiida-pseudo v{__version__}'
 
     if configuration not in SsspFamily.valid_configurations:
         echo.echo_critical(f'{version} {functional} {protocol} is not a valid SSSP configuration')
@@ -216,8 +218,9 @@ def cmd_install_sssp(version, functional, protocol, download_only, traceback):
         filepath_archive = dirpath / 'archive.tar.gz'
         filepath_metadata = dirpath / 'metadata.json'
 
-        download_sssp(configuration, filepath_archive, filepath_metadata, traceback)
+        patch_version = download_sssp(configuration, filepath_archive, filepath_metadata, traceback)
 
+        description = f'SSSP v{patch_version} {functional} {protocol} installed with aiida-pseudo v{__version__}'
         description += f'\nArchive pseudos md5: {md5_file(filepath_archive)}'
         description += f'\nPseudo metadata md5: {md5_file(filepath_metadata)}'
 
--- a/aiida_pseudo/cli/install.py
+++ b/aiida_pseudo/cli/install.py
@@ -3,7 +3,9 @@
 import json
 import pathlib
 import shutil
+from sys import version
 import tempfile
+import yaml
 
 import click
 import requests
@@ -108,9 +110,27 @@ def download_sssp(
     from aiida_pseudo.groups.family import SsspFamily
     from .utils import attempt
 
-    url_sssp_base = 'https://legacy-archive.materialscloud.org/file/2018.0001/v4/'
-    url_archive = f"{url_sssp_base}/{SsspFamily.format_configuration_filename(configuration, 'tar.gz')}"
-    url_metadata = f"{url_sssp_base}/{SsspFamily.format_configuration_filename(configuration, 'json')}"
+    url_prefix = 'https://staging-invenio.materialscloud.org/record/file?filename='
+    url_suffix = '&parent_id=19'
+
+    url_version_yaml = f"{url_prefix}versions.yaml{url_suffix}"
+
+    with attempt('downloading patch versions information... ', include_traceback=traceback):
+        response = requests.get(url_version_yaml)
+        response.raise_for_status()
+        version_mapping = yaml.load(response.content, Loader=yaml.FullLoader)
+
+    echo.echo_info(f'Latest patch version found: {version_mapping[configuration.version]}')
+
+    archive_filename = SsspFamily.format_configuration_filename(
+        configuration, 'tar.gz', version_mapping[configuration.version]
+    )
+    metadata_filename = SsspFamily.format_configuration_filename(
+        configuration, 'json', version_mapping[configuration.version]
+    )
+
+    url_archive = f"{url_prefix}{archive_filename}{url_suffix}"
+    url_metadata = f"{url_prefix}{metadata_filename}{url_suffix}"
 
     with attempt('downloading selected pseudopotentials archive... ', include_traceback=traceback):
         response = requests.get(url_archive)
--- a/aiida_pseudo/cli/install.py
+++ b/aiida_pseudo/cli/install.py
@@ -356,7 +356,7 @@ def cmd_install_pseudo_dojo(
             adjusted_cutoffs = {}
             for stringency, str_cutoffs in cutoffs.items():
                 adjusted_cutoffs[stringency] = []
-                max_cutoff_wfc = max([element_cutoffs['cutoff_wfc'] for element_cutoffs in str_cutoffs.values()])
+                max_cutoff_wfc = max([cutoffs['cutoff_wfc'] for cutoffs in str_cutoffs.values()])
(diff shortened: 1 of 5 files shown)


mbercx:
@sphuber the links in this PR are still for the staging version of the archive, since we wanted to test this before @vgranata puts this in production.

If you agree with the implementation, I'll ask @vgranata to move forward, then update the final links and fix the tests.

aiida_pseudo/cli/install.py
```diff
@@ -108,9 +110,27 @@ def download_sssp(
...
-    url_metadata = f"{url_sssp_base}/{SsspFamily.format_configuration_filename(configuration, 'json')}"
+    url_prefix = 'https://staging-invenio.materialscloud.org/record/file?filename='
+    url_suffix = '&parent_id=19'
+
+    url_version_yaml = f"{url_prefix}versions.yaml{url_suffix}"
+
+    with attempt('downloading patch versions information... ', include_traceback=traceback):
+        response = requests.get(url_version_yaml)
+        response.raise_for_status()
+        version_mapping = yaml.load(response.content, Loader=yaml.FullLoader)
```
sphuber, line 121: Probably want to use `SafeLoader`. I take it the actual version mapping should just contain text and base types, so the `SafeLoader` should be able to load it. This will protect us from arbitrary code execution should the URL ever be hijacked.
```suggestion
        version_mapping = yaml.load(response.content, Loader=yaml.SafeLoader)
```

aiida_pseudo/cli/install.py
```diff
@@ -108,9 +110,27 @@ def download_sssp(
...
-    url_sssp_base = 'https://legacy-archive.materialscloud.org/file/2018.0001/v4/'
-    url_archive = f"{url_sssp_base}/{SsspFamily.format_configuration_filename(configuration, 'tar.gz')}"
-    url_metadata = f"{url_sssp_base}/{SsspFamily.format_configuration_filename(configuration, 'json')}"
+    url_prefix = 'https://staging-invenio.materialscloud.org/record/file?filename='
+    url_suffix = '&parent_id=19'
+
+    url_version_yaml = f"{url_prefix}versions.yaml{url_suffix}"
+
+    with attempt('downloading patch versions information... ', include_traceback=traceback):
+        response = requests.get(url_version_yaml)
```
sphuber, line 119:
```suggestion
    url_template = 'https://archive.materialscloud.org/record/file?filename={filename}&parent_id=19'

    with attempt('downloading patch versions information... ', include_traceback=traceback):
        response = requests.get(url_template.format(filename='versions.yaml'))
```

aiida_pseudo/cli/install.py
```diff
@@ -108,9 +110,27 @@ def download_sssp(
...
+
+    archive_filename = SsspFamily.format_configuration_filename(
+        configuration, 'tar.gz', version_mapping[configuration.version]
+    )
+    metadata_filename = SsspFamily.format_configuration_filename(
+        configuration, 'json', version_mapping[configuration.version]
+    )
+
+    url_archive = f"{url_prefix}{archive_filename}{url_suffix}"
+    url_metadata = f"{url_prefix}{metadata_filename}{url_suffix}"
```
sphuber, line 133:
```suggestion
    url_archive = url_template.format(filename=archive_filename)
    url_metadata = url_template.format(filename=metadata_filename)
```

aiida_pseudo/cli/install.py
```diff
@@ -108,9 +110,27 @@ def download_sssp(
...
+
+    with attempt('downloading patch versions information... ', include_traceback=traceback):
+        response = requests.get(url_version_yaml)
+        response.raise_for_status()
+        version_mapping = yaml.load(response.content, Loader=yaml.FullLoader)
+
+    echo.echo_info(f'Latest patch version found: {version_mapping[configuration.version]}')
+
+    archive_filename = SsspFamily.format_configuration_filename(
+        configuration, 'tar.gz', version_mapping[configuration.version]
```
sphuber, line 126: Why did you add the argument `patch_version` to `format_configuration_filename` if you are not using it here. What will the output of `version_mapping[configuration.version]` be? Does that contain the patch version? Think it would be good to add a comment above line 121 with an example of the structure of the version mapping yaml returned by the first request.

aiida_pseudo/groups/family/sssp.py
```diff
@@ -1,7 +1,10 @@
 # -*- coding: utf-8 -*-
 """Subclass of ``PseudoPotentialFamily`` designed to represent an SSSP configuration."""
 from collections import namedtuple
-from typing import Sequence
+from typing import Sequence, Optional
+
+from aiida.cmdline import params
+from wrapt.wrappers import PartialCallableObjectProxy
```
sphuber, line 7: don't think you are really using this are you?

aiida_pseudo/groups/family/sssp.py
```diff
@@ -50,15 +53,17 @@ def format_configuration_label(cls, configuration: SsspConfiguration) -> str:
         )
 
     @classmethod
-    def format_configuration_filename(cls, configuration: SsspConfiguration, extension: str) -> str:
+    def format_configuration_filename(cls, configuration: SsspConfiguration, extension: str, patch_version: Optional[str] = None) -> str:
```
sphuber, line 56: Not sure if it is a good idea to keep the patch version as a separate argument. Shouldn't we try to integrate this in the `SsspConfiguration`. It already contains a `version` attribute. If that is `1.2.3` than the patch version is `3` by definition. So to have a separate argument that can cause inconsistencies seems weird.

sphuber (changes requested):
Thanks @mbercx . Have some suggestions and questions

aiida_pseudo/groups/family/sssp.py
mbercx: Oops, I first made some different adjustments, but forgot to remove these. (Also forgot to set up the pre-commit in the new dev environment I set up for `aiida-pseudo` 🤦 ) Will fix! 👍

aiida_pseudo/cli/install.py
mbercx: I am using it, no? `version_mapping[configuration.version]` will indeed return the patch version, but probably best to just grab this once and put it in a `patch_version` variable.

mbercx:
Thanks for the review, @sphuber! I made some first corrections, but have a look at [this comment](https://github.com/aiidateam/aiida-pseudo/pull/104#discussion_r686941153) and see if you agree with my reasoning here before I continue.

mbercx:
image

```console
(aiida-pseudo) mbercx@theospc46:~/envs/aiida-pseudo/code/aiida-pseudo$ git commit -a
Fix double quoted strings................................................Passed
Fix End of Files.........................................................Passed
Fix python encoding pragma...............................................Passed
Mixed line ending........................................................Passed
Trim Trailing Whitespace.................................................Passed
flynt....................................................................Passed
yapf.....................................................................Passed
pylint...................................................................Passed
pydocstyle...............................................................Passed
[fix/13/materialscloud-links 8874770] Apply more corrections from prime ape instance
 2 files changed, 4 insertions(+), 6 deletions(-)
```

aiida_pseudo/cli/install.py
sphuber: I see. Maybe would be useful to use the keyword to make it clearer. The fact that it also isn't very clear, is because it is not immediately clear what the `version_mapping` contains. If you add a comment there, it makes everything easier to comprehend

aiida_pseudo/groups/family/sssp.py
```diff
@@ -50,15 +50,19 @@ def format_configuration_label(cls, configuration: SsspConfiguration) -> str:
         )
 
     @classmethod
-    def format_configuration_filename(cls, configuration: SsspConfiguration, extension: str) -> str:
+    def format_configuration_filename(
+        cls, configuration: SsspConfiguration, extension: str, patch_version: Optional[str] = None
```
sphuber, line 54: Please update the docstring to explain what the `patch_version` looks like and what its purpose is.

aiida_pseudo/groups/family/sssp.py
sphuber: > Thanks for the review, @sphuber! I made some first corrections, but have a look at [this comment](https://github.com/aiidateam/aiida-pseudo/pull/104#discussion_r686941153) and see if you agree with my reasoning here before I continue.

I can remember the discussion about not fixing the patch number explicitly as that would require updates of the plugin if a new patch is released of the family, but I am having difficulty seeing how best to integrate this limitation. Can you share the content of the YAML mapping? Than I can form a better idea of what to do

aiida_pseudo/groups/family/sssp.py
mbercx: aye-aye-captain

aiida_pseudo/groups/family/sssp.py
mbercx: Sure! It's a very simple mapping of the minor version to the latest patch version:

```yaml
---
# This maps each minor version to the latest patch version
# corresponding to that minor version
'1.0': '1.0'
'1.1': '1.1.2'
```

aiida_pseudo/cli/install.py
```diff
@@ -108,9 +109,25 @@ def download_sssp(
     from aiida_pseudo.groups.family import SsspFamily
     from .utils import attempt
 
-    url_sssp_base = 'https://legacy-archive.materialscloud.org/file/2018.0001/v4/'
-    url_archive = f"{url_sssp_base}/{SsspFamily.format_configuration_filename(configuration, 'tar.gz')}"
-    url_metadata = f"{url_sssp_base}/{SsspFamily.format_configuration_filename(configuration, 'json')}"
+    url_template = 'https://archive.materialscloud.org/record/file?filename={filename}&parent_id=19'
```
sphuber, line 112: Thanks for the changes. Everything looks good to me now, except I still have a question about this magic `parent_id=19`. I have been trying to figure out how it relates to the main archive entry. I take it it refers to the actual archive record: https://archive.materialscloud.org/record/2021.76
As you can see, there in the URL it uses an ID that seems based on the year and some auto-incrementing number. But this ID changes if a new version of the SSSP record is created. The question is whether the `parent_id=19` is persistent and doesn't change. I.e., when a `v8` will be released with the next patch version of `1.1`, i.e. `1.1.3`, will the URL still work?

sphuber:
Fixed the CI issues in #105

aiida_pseudo/cli/install.py
vgranata: The parent_id is persistent within versions of the same record, meaning that it does not change if a new version of a record is created. 
If you create a new SSSP record that is not a new version of https://archive.materialscloud.org/record/2021.76 then the parent_id will be different.

sphuber: OK Thanks @vgranata that should be fine. Then we can almost merge this @mbercx . Only thing is that we should probably add the patch version to the description of the group. Otherwise it will only be possible to know what patch version a certain family is by looking at the md5 of the tarball that was downloaded, which is contained in the description. I think it would make sense to store the patch at the very least in the description of the family. And then the tests should be fixed of course, because they are failing

mbercx:
@sphuber I've removed the `md5` hashes and the version from the `test_sssp_install` test, since these will depend on the patch version. I.e. if this is updated on the archive, the test will start to fail.

aiida_pseudo/cli/install.py
```diff
@@ -104,13 +105,30 @@ def download_sssp(
     :param filepath_archive: absolute filepath to write the pseudopotential archive to.
     :param filepath_metadata: absolute filepath to write the metadata file to.
     :param traceback: boolean, if true, print the traceback when an exception occurs.
+    :return: Latest patch version of the requested minor version
```
sphuber, line 108: Now just need to update the return type of this function as well and than this is good to merge

mbercx: Righto! Missed that one, thanks

sphuber: More a sign that we should probably at some point at `mypy` to the pre-commit. On the one hand I like added typing, but on the other hand `mypy` has been given so much trouble that sometimes it annoys me to no end. :rock: me :hardplace:

sphuber (approved):

sphuber:
Thanks @mbercx

PR #110, opened 2021-12-08
Replace SSSP link to legacy MC Archive

### LINKED ISSUE #109: Remove the usage of legacy link for SSSP in tests
One unit test still attempts to install the SSSP from the legacy server of the MC archive which has now officiallly been decommissioned.

sphuber:
Fixes #109 

The `cli.test_install:test_install_family_url` test was using a link to
the legacy version of the Materials Cloud Archive which has now been
taken offline, causing the test to fail. The link is updated to point to
the new server.


sphuber:
Alright @mbercx , took three more problems to be solved to solve the original problem, but with this, the CI is running again

PR #132, opened 2022-10-09
CLI: filter cutoffs from JSON file for family cutoffs set

### LINKED ISSUE #93: CLI: `family cutoffs set` does not accept SSSP metadata

mbercx:
Fixes #93 

Currently the `family cutoffs set` command requires that the JSON file that contains the cutoffs _only_ has the cutoff keys specified for each element. This is because the cutoffs are validated by the `RecommendedCutoffMixin.validate_cutoffs()` method, which doesn't allow for extraneous keys.

However, the  `.json`  file could have been adapted by the user from e.g. the SSSP `metadata.json` or some other dictionary that also contains other keys. As long as the cutoff keys are defined, we shouldn't raise an error just because there are other keys present.

Here we first filter the data loaded from the JSON file for the cutoffs, before passing the dictionary to the `set_cutoffs` method.

### DIFF
--- a/src/aiida_pseudo/cli/family.py
+++ b/src/aiida_pseudo/cli/family.py
@@ -90,7 +90,9 @@ def cmd_family_cutoffs_set(family, cutoffs, stringency, unit):  # noqa: D301
         try:
             cutoffs_dict[element] = {'cutoff_wfc': values['cutoff_wfc'], 'cutoff_rho': values['cutoff_rho']}
         except KeyError as exception:
-            raise click.BadParameter(f'`{cutoffs.name}` is missing cutoffs for element: {element}') from exception
+            raise click.BadParameter(
+                f'`{cutoffs.name}` is missing cutoffs for element `{element}`: {exception}', param_hint='CUTOFFS'
+            ) from exception
 
     try:
         family.set_cutoffs(cutoffs_dict, stringency, unit=unit)
--- a/src/aiida_pseudo/cli/family.py
+++ b/src/aiida_pseudo/cli/family.py
@@ -85,8 +85,15 @@ def cmd_family_cutoffs_set(family, cutoffs, stringency, unit):  # noqa: D301
     except ValueError as exception:
         raise click.BadParameter(f'`{cutoffs.name}` contains invalid JSON: {exception}', param_hint='CUTOFFS')
 
+    cutoffs_dict = {}
+    for element, values in data.items():
+        try:
+            cutoffs_dict[element] = {'cutoff_wfc': values['cutoff_wfc'], 'cutoff_rho': values['cutoff_rho']}
+        except KeyError as exception:
+            raise click.BadParameter(f'`{cutoffs.name}` is missing cutoffs for element: {element}') from exception
+
     try:
-        family.set_cutoffs(data, stringency, unit=unit)
+        family.set_cutoffs(cutoffs_dict, stringency, unit=unit)
     except ValueError as exception:
         raise click.BadParameter(f'`{cutoffs.name}` contains invalid cutoffs: {exception}', param_hint='CUTOFFS')
 
--- a/tests/cli/test_family.py
+++ b/tests/cli/test_family.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 # pylint: disable=unused-argument,redefined-outer-name
 """Tests for the command `aiida-pseudo family`."""
+from copy import deepcopy
 import json
 
 from aiida.orm import Group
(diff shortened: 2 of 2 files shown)


mbercx:
@sphuber while working myself back into https://github.com/aiidateam/aiida-pseudo/pull/94 I decided to separate the work into different PR's since not everything is related to the same topic and it'll be easier to rebase/adapt/review.

src/aiida_pseudo/cli/family.py
```diff
@@ -85,8 +85,15 @@ def cmd_family_cutoffs_set(family, cutoffs, stringency, unit):  # noqa: D301
     except ValueError as exception:
         raise click.BadParameter(f'`{cutoffs.name}` contains invalid JSON: {exception}', param_hint='CUTOFFS')
 
+    cutoffs_dict = {}
+    for element, values in data.items():
+        try:
+            cutoffs_dict[element] = {'cutoff_wfc': values['cutoff_wfc'], 'cutoff_rho': values['cutoff_rho']}
+        except KeyError as exception:
+            raise click.BadParameter(f'`{cutoffs.name}` is missing cutoffs for element: {element}') from exception
```
sphuber, line 93:
```suggestion
            raise click.BadParameter(f'`{cutoffs.name}` is missing cutoffs for element `{element}`: {exception}', param_hint='CUTOFFS') from exception
```

tests/cli/test_family.py
```diff
@@ -46,6 +49,16 @@ def test_family_cutoffs_set(run_cli_command, get_pseudo_family, generate_cutoffs
     assert sorted(family.get_cutoff_stringencies()) == sorted(['low', 'normal', 'high'])
     assert family.get_cutoffs(stringency) == cutoffs_dict[stringency]
 
+    # Invalid cutoffs structure - Should pass fine
```
sphuber, line 52:
```suggestion
    # Additional keys in the cutoffs should be accepted and simply ignored
```

src/aiida_pseudo/cli/family.py
```diff
@@ -85,8 +85,15 @@ def cmd_family_cutoffs_set(family, cutoffs, stringency, unit):  # noqa: D301
     except ValueError as exception:
         raise click.BadParameter(f'`{cutoffs.name}` contains invalid JSON: {exception}', param_hint='CUTOFFS')
 
+    cutoffs_dict = {}
+    for element, values in data.items():
+        try:
+            cutoffs_dict[element] = {'cutoff_wfc': values['cutoff_wfc'], 'cutoff_rho': values['cutoff_rho']}
+        except KeyError as exception:
```
sphuber, line 92: I don't think the original exception message will be included this way, will it? So if this exception is raised, it is not clear which key is missing. Of course people could look at the docstring to figure out that `cutoff_wfc` and `cutoff_rho` must be used, but maybe just including the original error might be easier.

sphuber (changes requested):
Thanks @mbercx just some minor comments

src/aiida_pseudo/cli/family.py
mbercx: Indeed, much clearer with the change in your other comment:
```console
$ aiida-pseudo family cutoffs set -s test -u Ry test-family metadata.json 
Try 'aiida-pseudo family cutoffs set -h' for help.

Error: Invalid value for CUTOFFS: `metadata.json` is missing cutoffs for element `Ag`: 'cutoff_wfc'
```

sphuber (approved):

sphuber:
Thanks @mbercx

PR #134, opened 2022-10-09
CLI: block family cutoffs set for established families

mbercx:
One of the motivations of having dedicated and automated CLI commands to install established pseudpotential families like the SSSP or the Pseudo-dojo normconserving pseudopotentials is to try and prevent that the pseudopotentials or recommended cutoffs deviate from the official ones. However, the `family cutoffs set` command currently allows the user to change the recommended cutoffs easily through the CLI.

In order to prevent this, we block usage of the command for `SsspFamily` and `PseudoDojoFamily`'s, by adding an `exclude` input argument to the `PseudoPotentialFamilyParam`, where the entry points of the classes that should be disallowed can be provided. This is then used to adapt the option decorator for the `PSEUDO_POTENTIAL_FAMILY` input argument of `family cutoffs set`.

Note that the user _can_ still install e.g. the SSSP with their own recommended cutoffs using the `install family` command as a `CutoffsPseudoPotentialFamily`. However, the `SsspFamily` class is reserved for pseudopotentials installed with the automated install command.

### DIFF
--- a/docs/source/design.rst
+++ b/docs/source/design.rst
@@ -103,9 +103,9 @@ A pseudopotential family is only considered to be *established* when it has a co
 
 Only a pseudopotential family that is *established* will receive support for automated installs with its own class (e.g. :py:class:`~aiida_pseudo.groups.family.SsspFamily`/:py:class:`~aiida_pseudo.groups.family.PseudoDojoFamily`) and command line interface (CLI) commands (e.g. ``install sssp``/``install pseudo-dojo``).
 To make sure these families represent the official ones, they can only be installed with their supported CLI commands, and there are strict checks on the format of these files to make sure they correspond to the official ones.
-Based on the same principle of preserving the integrity of these established pseudopotentials, the ``family cutoffs set`` command can not be used to set the recommended cutoffs of an established family.
+Based on the same principle of preserving the integrity of these established pseudopotentials, the ``family cutoffs set`` command cannot be used to set the recommended cutoffs of an established family.
 
-If users want to install the a set of *non-established* pseudopotentials and configure their recommended cutoffs, they have to install them from the archive using ``install family`` as a :py:class:`~aiida_pseudo.groups.family.CutoffsPseudoPotentialFamily` as described in the :ref:`corresponding how-to section <how-to:install_archive>`.
+To install a set of *non-established* pseudopotentials and configure their recommended cutoffs, install them from the archive using ``install family`` as a :py:class:`~aiida_pseudo.groups.family.CutoffsPseudoPotentialFamily` as described in the :ref:`corresponding how-to section <how-to:install_archive>`.
 
 Recommended cutoffs
 ===================
--- a/docs/source/design.rst
+++ b/docs/source/design.rst
@@ -11,14 +11,14 @@ The two concepts are further explained below, especially focusing on how they ar
 Pseudopotentials
 ================
 
-Pseudopotentials are implemented as `data plugins <https://aiida-core.readthedocs.io/en/latest/topics/data_types.html#creating-a-data-plugin>`_, and the base class is ``PseudoPotentialData``.
+Pseudopotentials are implemented as `data plugins <https://aiida-core.readthedocs.io/en/latest/topics/data_types.html#creating-a-data-plugin>`_, and the base class is :py:class:`~aiida_pseudo.data.pseudo.PseudoPotentialData`.
 Pseudopotentials are assumed to be defined by a single file on disk and represent a single chemical element.
 As such, each pseudopotential node, *has* to have two attributes: the md5 of the file that it represents and the symbol of chemical element that the pseudopotential describes.
-The latter follows IUPAC naming conventions as used in the module ``aiida.common.constants.elements`` of ``aiida-core``.
+The latter follows IUPAC naming conventions defined as ``elements`` in the module ``aiida.common.constants`` of ``aiida-core``.
 
-The ``PseudoPotentialData`` functions as a base class and does not represent an actual existing pseudopotential format.
-It *is* possible to create a family of ``PseudoPotentialData`` nodes, but since the element cannot be parsed from the file content itself, it will have to be done from the *filename*, so the filename *has* to have the format ``CHEMICAL_SYMBOL.EXTENSION``, for example ``Ar.pseudo``.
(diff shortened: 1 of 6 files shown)


docs/source/design.rst
```diff
@@ -90,17 +90,29 @@ If you have a folder on disk that contains various pseudopotentials for differen
...
+
+Established families
+--------------------
+
+When it comes to pseudopotential families, ``aiida-pseudo`` makes a clear distinction between families that are *established* and those that are not.
+A pseudopotential family is only considered to be *established* when it has a comprehensive set of rigorously tested pseudopotentials with convergence tests, both of which have been published and are publicly available.
+
+Only a pseudopotential family that is *established* will receive support for automated installs with its own class (e.g. :py:class:`~aiida_pseudo.groups.family.SsspFamily`/:py:class:`~aiida_pseudo.groups.family.PseudoDojoFamily`) and command line interface (CLI) commands (e.g. ``install sssp``/``install pseudo-dojo``).
+To make sure these families represent the official ones, they can only be installed with their supported CLI commands, and there are strict checks on the format of these files to make sure they correspond to the official ones.
+Based on the same principle of preserving the integrity of these established pseudopotentials, the ``family cutoffs set`` command can not be used to set the recommended cutoffs of an established family.
```
sphuber, line 106:
```suggestion
Based on the same principle of preserving the integrity of these established pseudopotentials, the ``family cutoffs set`` command cannot be used to set the recommended cutoffs of an established family.
```

docs/source/design.rst
```diff
@@ -90,17 +90,29 @@ If you have a folder on disk that contains various pseudopotentials for differen
...
+
+If users want to install the a set of *non-established* pseudopotentials and configure their recommended cutoffs, they have to install them from the archive using ``install family`` as a :py:class:`~aiida_pseudo.groups.family.CutoffsPseudoPotentialFamily` as described in the :ref:`corresponding how-to section <how-to:install_archive>`.
```
sphuber, line 108:
```suggestion
To install a set of *non-established* pseudopotentials and configure their recommended cutoffs, install them from the archive using ``install family`` as a :py:class:`~aiida_pseudo.groups.family.CutoffsPseudoPotentialFamily` as described in the :ref:`corresponding how-to section <how-to:install_archive>`.
```

src/aiida_pseudo/cli/params/types.py
```diff
@@ -54,13 +55,34 @@ def complete(self, _, incomplete):
 
 
 class PseudoPotentialFamilyParam(GroupParamType):
-    """Parameter type for `click` commands to define an instance of a `PseudoPotentialFamily`."""
+    """Parameter type for ``click`` commands to define an instance of a ``PseudoPotentialFamily``."""
 
     name = 'pseudo_family'
 
+    def __init__(self, exclude: typing.Optional[typing.List[str]] = None, **kwargs):
```
sphuber, line 62:
```suggestion
    def __init__(self, exclude: list[str] | None = None, **kwargs):
```
You can use this notation if you put `from __future__ import annotations` at the top of the file (just below the docstring). We should start to use this everywhere, it is so much more readable.

src/aiida_pseudo/cli/params/types.py
```diff
@@ -54,13 +55,34 @@ def complete(self, _, incomplete):
...
+        """Construct the parameter.
+
+        :param exclude: an optional list of values that should be considered invalid and will raise ``BadParameter``.
```
sphuber, line 65: The word `exclude` seems a bit soft to me and doesn't immediately suggest that these will incur exceptions. Would `blacklist` maybe be better? As these are values that are actually blacklisted and not accepted?
```suggestion
        :param exclude: an optional list of values that should be considered invalid and will raise ``BadParameter``.
```

tests/cli/test_family.py
```diff
@@ -80,6 +83,30 @@ def test_family_cutoffs_set_unit(run_cli_command, get_pseudo_family, generate_cu
     assert family.get_cutoffs_unit() == unit
 
 
+@pytest.mark.parametrize(
+    'family_cls,label', [(SsspFamily, 'SSSP/1.1/PBE/efficiency'),
+                         (PseudoDojoFamily, 'PseudoDojo/0.4/PBE/SR/standard/psp8')]
+)
+@pytest.mark.usefixtures('clear_db')
+def test_family_cutoffs_set_established(
```
sphuber, line 91: Is there still a test for a family type that is actually accepted?

sphuber (changes requested):
Thanks @mbercx . Very nice and clear. Just some minor comments

src/aiida_pseudo/cli/params/types.py
mbercx: Would 
```python
def convert(self, value, param, ctx) -> typing.Union[pathlib.Path, bytes]:
```

Then also become
```python
def convert(self, value, param, ctx) -> pathlib.Path | bytes:
```

Do you have any good tutorial / reference on this?

src/aiida_pseudo/cli/params/types.py
mbercx: Sure, `blacklist` also works. Or maybe `invalid`? Trying to use one of the "excluded" family types will indeed fail:

```console
$ aiida-pseudo family cutoffs set SSSP/1.1/PBE/efficiency metadata.json 
Try 'aiida-pseudo family cutoffs set -h' for help.

Error: Invalid value for 'FAMILY': The value `SsspFamily<SSSP/1.1/PBE/efficiency>` is not allowed for this parameter.
```

tests/cli/test_family.py
mbercx: Yes, `test_family_cutoffs_set` is still there!

src/aiida_pseudo/cli/params/types.py
sphuber: yes, the `|` operator functions the same as `typing.Union`. Best reference are official docs: https://docs.python.org/3/library/typing.html

src/aiida_pseudo/cli/params/types.py
sphuber: I think `blacklist` is more correct since they aren't so much *invalid* as they are simply not allowed for this particular instance of the parameter type.

sphuber (approved):

sphuber:
Beautiful, all good, thanks @mbercx

PR #135, opened 2022-10-10
CLI: add install `--from-download` for sssp and pseudo-dojo

### LINKED ISSUE #92: `install family`: allow installation of sssp and pseudo-dojo

mbercx:
Fixes #92 

Our workaround for installing an "established" family like a `SsspFamily` or `PseudoDojoFamily` required using the `install family` command, but this is explicitly forbidden since we do not want the user to be able to install an arbitrary set of pseudopotentials as one of these classes. This means there is currently no way to install the archive and metadata obtained with the `--download-only` option of the automated install commands as a proper `SsspFamily` or `PseudoDojoFamily`.

Here, we add the `--from-download` option to the automated install commands to allow the user to install the downloaded pseudopotentials and set the recommended cutoffs automatically for an `SsspFamily` or `PseudoDojoFamily`. The `--download-only` option now also produces a single `.aiida_pseudo` archive that contains:

* The archive containing the pseudos
* The corresponding metadata
* The configuration in JSON format

The user can pass such an archive to the `from-download` option to install the corresponding pseudopotential family.

### DIFF
--- a/docs/source/howto.rst
+++ b/docs/source/howto.rst
@@ -28,12 +28,16 @@ Similar to the `SSSP`_ pseudopotential family, the options can be used to set ve
 Moreover, the format of the pseudopotentials can be specified, as well as the default stringency.
 Use ``aiida-pseudo install pseudo-dojo --help`` for more information.
 
+.. note::
+
+    In case you are unable to use these automated commands because of connection issues, please consult the :ref:`troubleshooting section <troubleshooting:automated-fail>`.
+
 .. _how-to:install_archive:
 
 Installing from archive or folder
 =================================
 
-In case the automated install commands fail, or the pseudopotential family you want to use is not supported, you can install the pseudopotential family manually with the following command:
+In case the pseudopotential family you want to use is not supported, you can install the pseudopotential family manually with the following command:
 
 .. code-block:: console
 
@@ -53,7 +57,7 @@ If this fails, you can specify the format manually with the ``--archive-format/-
 
     $ aiida-pseudo install family <ARCHIVE> <LABEL> -f gztar
 
-The valid archive formats are those [defined by the ``shutil.unpack_archive`` function](https://docs.python.org/3/library/shutil.html#shutil.unpack_archive) of the standard Python library.
+The valid archive formats are those defined by the `shutil.unpack_archive <https://docs.python.org/3/library/shutil.html#shutil.unpack_archive>`_ function of the standard Python library.
 
 Pseudopotential format
 ----------------------
@@ -93,6 +97,7 @@ The available pseudopotential family classes can be listed with the command:
 .. important::
 
     The ``pseudo.family.sssp`` and ``pseudo.family.pseudo_dojo`` family types are blacklisted since they have their own :ref:`dedicated install commands <how-to:install_automated>` in ``aiida-pseudo install sssp`` and ``aiida-pseudo install pseudo-dojo``, respectively.
+    In case you are unable to use these commands because of connection issues, please consult the :ref:`troubleshooting section <troubleshooting:automated-fail>`.
 
 Adding recommended cutoffs
 --------------------------
--- a/docs/source/troubleshooting.rst
+++ b/docs/source/troubleshooting.rst
@@ -5,9 +5,27 @@
 Troubleshooting
 ###############
 
+.. _troubleshooting:automated-fail:
+
 The automated install commands fail
 ===================================
 
 These failures are often due to unstable internet connections causing the download of the pseudopotential archive from the web to fail.
-In this case, it is possible to install the family manually from an archive that is already available on the local file system.
-You can read more on this in the :ref:`how-to section <how-to:install_archive>`.
+In this case, it is possible to download an ``.aiida_pseudo`` archive of the established family, transfer them to the machine where the pseudopotentials need to be installed and install directly from the archive.
+To download the archive, use the ``--download-only`` option:
+
+.. code-block:: console
+
+    $ aiida-pseudo install sssp --download-only
+    Report: downloading patch versions information...  [OK]
+    Report: downloading selected pseudopotentials archive...  [OK]
+    Report: downloading selected pseudopotentials metadata...  [OK]
+    Success: Pseudopotential archive written to: SSSP_1.1_PBE_efficiency.aiida_pseudo
+
+The downloaded archive can be installed through the ``--from-download`` option of the corresponding install command:
+
+.. code-block:: console
+
+    $ aiida-pseudo install sssp --from-download SSSP_1.1_PBE_efficiency.aiida_pseudo
+    Report: unpacking archive and parsing pseudos...  [OK]
+    Success: installed `SSSP/1.1/PBE/efficiency` containing 85 pseudopotentials.
--- a/docs/source/troubleshooting.rst
+++ b/docs/source/troubleshooting.rst
@@ -11,7 +11,7 @@ The automated install commands fail
 ===================================
 
 These failures are often due to unstable internet connections causing the download of the pseudopotential archive from the web to fail.
-In this case, it is possible to download an ``.aiida_pseudo`` archive of the established family, transfer them to the machine where the pseudopotentials need to be installed and install directly from the archive.
+In this case, it is possible to download an ``.aiida_pseudo`` archive of the established family on a machine with a stable internet connection, transfer it to the machine where the pseudopotentials need to be installed and install directly from the archive.
 To download the archive, use the ``--download-only`` option:
 
 .. code-block:: console
--- a/src/aiida_pseudo/cli/install.py
+++ b/src/aiida_pseudo/cli/install.py
@@ -167,7 +167,8 @@ def install_sssp(
     # pylint: disable=too-many-locals
     from aiida.orm import Group
 
-    from ..groups.family.sssp import SsspFamily
+    from aiida_pseudo.groups.family.sssp import SsspFamily
+
     from .utils import attempt, create_family_from_archive
 
     with open(filepath_metadata, 'rb') as handle:
@@ -223,15 +224,13 @@ def cmd_install_sssp(version, functional, protocol, download_only, from_download
 
     configuration = SsspConfiguration(version, functional, protocol)
 
-    if configuration not in SsspFamily.valid_configurations:
-        echo.echo_critical(f'{version} {functional} {protocol} is not a valid SSSP configuration')
-
     with tempfile.TemporaryDirectory() as dirpath:
 
         dirpath = pathlib.Path(dirpath)
 
         filepath_archive = dirpath / 'archive.tar.gz'
         filepath_metadata = dirpath / 'metadata.json'
+        filepath_configuration = dirpath / 'configuration.json'
 
         if from_download is not None:
 
@@ -239,21 +238,24 @@ def cmd_install_sssp(version, functional, protocol, download_only, from_download
             with tarfile.open(tarball_path, 'r') as handle:
                 handle.extractall(dirpath)
 
-            filepath_configuration = dirpath / 'configuration.json'
+            for filepath in (filepath_archive, filepath_metadata, filepath_configuration):
+                if not filepath.exists():
+                    echo.echo_critical(
+                        f'archive `{from_download}` did not contain required file `{filepath_archive.name}`'
+                    )
 
             with filepath_configuration.open('r') as handle:
                 configuration = SsspConfiguration(**json.load(handle))
-        else:
+
+        if configuration not in SsspFamily.valid_configurations:
+            echo.echo_critical(f'{version} {functional} {protocol} is not a valid SSSP configuration')
+
+        if not from_download:
             download_sssp(configuration, filepath_archive, filepath_metadata, traceback)
 
         label = SsspFamily.format_configuration_label(configuration)
 
-        if not download_only and QueryBuilder().append(SsspFamily, filters={'label': label}).first():
-            echo.echo_report(f'{SsspFamily.__name__}<{label}> is already installed')
-            sys.exit(1)
-
         if download_only:
-            filepath_configuration = dirpath / 'configuration.json'
 
             with filepath_configuration.open('w') as handle:
                 handle.write(json.dumps(configuration._asdict()))
@@ -268,6 +270,10 @@ def cmd_install_sssp(version, functional, protocol, download_only, from_download
                     tarball_file.add(filepath, filepath.name)
             echo.echo_success(f'Pseudopotential archive written to: {tarball_path.name}')
             return
+
+        if QueryBuilder().append(SsspFamily, filters={'label': label}).first():
+            echo.echo_report(f'{SsspFamily.__name__}<{label}> is already installed')
+            sys.exit(1)
 
         description = (
             f'SSSP v{configuration.version} {configuration.functional} {configuration.protocol} '
--- a/src/aiida_pseudo/cli/install.py
+++ b/src/aiida_pseudo/cli/install.py
@@ -2,8 +2,8 @@
 """Command to install a pseudo potential family."""
 import json
 import pathlib
-import shutil
 import sys
+import tarfile
 import tempfile
 import typing as t
 
@@ -17,7 +17,8 @@ from .params import options, types
 from .root import cmd_root
 
 if t.TYPE_CHECKING:
-    from aiida_pseudo.groups.family import SsspConfiguration
+    from aiida_pseudo.data.pseudo import PseudoPotentialData
+    from aiida_pseudo.groups.family import PseudoDojoConfiguration, PseudoDojoFamily, SsspConfiguration
 
 
 @cmd_root.group('install')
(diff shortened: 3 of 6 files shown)


mbercx:
@sphuber still have to update the documentation instructions. Here's a quick example of the usage:

```console
(aiida-pseudo) mbercx@theospc46:~/envs/aiida-pseudo/data$ aiida-pseudo install pseudo-dojo -x LDA -p standard -f upf --download-only
Report: downloading selected pseudopotentials archive...  [OK]
Report: downloading selected pseudopotentials metadata archive...  [OK]
Success: Pseudopotential archive written to: PseudoDojo_0.4_LDA_SR_standard_upf.aiida_pseudo
(aiida-pseudo) mbercx@theospc46:~/envs/aiida-pseudo/data$ ls
PseudoDojo_0.4_LDA_SR_standard_upf.aiida_pseudo
(aiida-pseudo) mbercx@theospc46:~/envs/aiida-pseudo/data$ aiida-pseudo install pseudo-dojo --from-download PseudoDojo_0.4_LDA_SR_standard_upf.aiida_pseudo 
Report: unpacking archive and parsing pseudos...  [OK]
Report: unpacking metadata archive and parsing metadata.../home/mbercx/envs/aiida-pseudo/code/aiida-pseudo/src/aiida_pseudo/groups/family/pseudo_dojo.py:278: UserWarning: filename Ge.djrepo_old does not have a supported extension. Skipping...
  warnings.warn(f'filename {filename} does not have a supported extension. Skipping...')
 [OK]
Success: installed `PseudoDojo/0.4/LDA/SR/standard/upf` containing 70 pseudopotentials
```

If we still agree on the `--download-only` and `--from-download` approach, I'll update the docs as well.

sphuber:
> If we still agree on the --download-only and --from-download approach, I'll update the docs as well.

Remind me, did we any alternatives? Did we figure out any obvious problems of this approach? If not, it seems reasonable to me.

mbercx:
I think your comment here presents some of the alternatives discussed: https://github.com/aiidateam/aiida-pseudo/pull/94#issuecomment-839145418. You can also read the two comments below regarding the choice to switch to `--from-download`.

My one idea was still to extend this `.aiida_pseudo` archive idea to installing other families, so it's easier for people to share pseudo families with specified cutoffs, but that's beyond the scope of this PR. If you think it's a good idea I'll open an issue for it.

sphuber:
> I think your comment here presents some of the alternatives discussed: [#94 (comment)](https://github.com/aiidateam/aiida-pseudo/pull/94#issuecomment-839145418). You can also read the two comments below regarding the choice to switch to `--from-download`.

I remember now. I was wondering if only accepting a tarball with everything was not too limited. But I think I was maybe getting ahead of myself. Maybe let's first merge this and see if people actually need it and whether they would require even more control.

> My one idea was still to extend this `.aiida_pseudo` archive idea to installing other families, so it's easier for people to share pseudo families with specified cutoffs, but that's beyond the scope of this PR. If you think it's a good idea I'll open an issue for it.

I like the concept, feature request would be good to discuss the details.

mbercx:
@sphuber docs should be updated! So this PR is ready for review.

tests/cli/test_install.py
```diff
@@ -297,7 +297,22 @@ def test_install_sssp_download_only_exists(run_monkeypatched_install_sssp, get_p
...
+@pytest.mark.parametrize('configuration', SsspFamily.valid_configurations)
+@pytest.mark.usefixtures('clear_db', 'chdir_tmp_path')
+def test_install_sssp_from_download(run_monkeypatched_install_sssp, configuration):
+    """Test the ``aiida-pseudo install sssp`` command with the ``--from-download`` option."""
+    options = [
+        '--download-only', '-v', configuration.version, '-x', configuration.functional, '-p', configuration.protocol
+    ]
+    result = run_monkeypatched_install_sssp(options=options)
+
+    options = ['--from-download', result.output.split(': ')[-1][:-1]]
```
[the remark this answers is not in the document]
sphuber: Same for the pseudodojo test

docs/source/troubleshooting.rst
```diff
@@ -5,9 +5,27 @@
...
 
+.. _troubleshooting:automated-fail:
+
 The automated install commands fail
 ===================================
 
 These failures are often due to unstable internet connections causing the download of the pseudopotential archive from the web to fail.
-In this case, it is possible to install the family manually from an archive that is already available on the local file system.
-You can read more on this in the :ref:`how-to section <how-to:install_archive>`.
+In this case, it is possible to download an ``.aiida_pseudo`` archive of the established family, transfer them to the machine where the pseudopotentials need to be installed and install from a directory.
```
sphuber, line 14:
```suggestion
In this case, it is possible to download an ``.aiida_pseudo`` archive of the established family, transfer them to the machine where the pseudopotentials need to be installed and install directly from the archive.
```

docs/source/troubleshooting.rst
```diff
@@ -5,9 +5,27 @@
...
+
+.. code-block:: console
+
+    $ aiida-pseudo install sssp --download-only
+    Report: downloading patch versions information...  [OK]
+    Report: downloading selected pseudopotentials archive...  [OK]
+    Report: downloading selected pseudopotentials metadata...  [OK]
+    Success: Pseudopotential archive written to: SSSP_1.1_PBE_efficiency.aiida_pseudo
+
+You can then transfer the downloaded archive to the machine where it should be installed use the ``--from-download`` option of the corresponding install command:
```
sphuber, line 25:
```suggestion
The downloaded archive can be installed through the ``--from-download`` option of the corresponding install command:
```

docs/source/howto.rst
```diff
@@ -53,7 +57,7 @@ If this fails, you can specify the format manually with the ``--archive-format/-
 
     $ aiida-pseudo install family <ARCHIVE> <LABEL> -f gztar
 
-The valid archive formats are those [defined by the ``shutil.unpack_archive`` function](https://docs.python.org/3/library/shutil.html#shutil.unpack_archive) of the standard Python library.
+The valid archive formats are those `defined by the shutil.unpack_archive function <https://docs.python.org/3/library/shutil.html#shutil.unpack_archive>`_ of the standard Python library.
```
sphuber, line 60:
```suggestion
The valid archive formats are those defined by the `shutil.unpack_archive` <https://docs.python.org/3/library/shutil.html#shutil.unpack_archive>`_ function of the standard Python library.
```

src/aiida_pseudo/cli/params/options.py
```diff
@@ -103,3 +103,12 @@
...
     )
 )
+
+FROM_DOWNLOAD = OverridableOption(
+    '--from-download',
+    type=click.STRING,
+    required=False,
+    default=None,
+    show_default=False,
+    help='Install the pseudpotential family from the archive and metadata downloaded with the `--download-only` option.'
```
sphuber, line 113:
```suggestion
    help='Install the pseudopotential family from an archive downloaded with the `--download-only` option.'
```

src/aiida_pseudo/cli/install.py
```diff
@@ -143,117 +140,215 @@ def download_sssp(
     return patch_version
 
 
-def download_pseudo_dojo(
-    configuration: SsspConfiguration,
+def install_sssp(
```
sphuber, line 143: Please add docstring for parameters

src/aiida_pseudo/cli/install.py
```diff
@@ -143,117 +140,215 @@ def download_sssp(
...
+
+    with attempt('downloading selected pseudopotentials metadata archive... ', include_traceback=traceback):
+        response = requests.get(url_metadata)
+        response.raise_for_status()
+        with open(filepath_metadata, 'wb') as handle:
+            handle.write(response.content)
+            handle.flush()
+
+
+def install_pseudo_dojo(
```
sphuber, line 289: Please add docstring for parameters

src/aiida_pseudo/cli/install.py
```diff
@@ -143,117 +140,215 @@ def download_sssp(
...
 
     configuration = SsspConfiguration(version, functional, protocol)
-    label = SsspFamily.format_configuration_label(configuration)
 
     if configuration not in SsspFamily.valid_configurations:
-        echo.echo_critical(f'{configuration} is not a valid configuration.')
-
-    if not download_only and QueryBuilder().append(SsspFamily, filters={'label': label}).first():
-        echo.echo_critical(f'{SsspFamily.__name__}<{label}> is already installed')
+        echo.echo_critical(f'{version} {functional} {protocol} is not a valid SSSP configuration')
```
sphuber, line 203: We should probably move this check to somewhere around after line 221 when the configuration from the archive is also created. Currently that is not validated I believe. Simple validation that the config in the archive is still valid

src/aiida_pseudo/cli/install.py
```diff
@@ -143,117 +140,215 @@ def download_sssp(
...
-                    # Cannot use ``pathlib.Path.rename`` because this will fail if it moves across file systems.
-                    shutil.move(filepath, filepath_target)
-                    echo.echo_success(f'`{filepath_target.name}` written to the current directory.')
+            filepath_configuration = dirpath / 'configuration.json'
+
+            with filepath_configuration.open('w') as handle:
+                handle.write(json.dumps(configuration._asdict()))
+
+            tarball_path = pathlib.Path.cwd() / f'{label}.aiida_pseudo'.replace('/', '_')
+
```
sphuber, line 237: The logic in this block is quite hard to follow. I think it will be made a lot easier if you simply take out the case for `download_only == True`. Take that out of the `with tempfile` context (or give it its own) and simply return after it is done.
```
if download_only:
    with tempfile.TemporaryDirectory() as dirpath:
         # logic to download
         return

# Here we now know we should install normally or `--from-download`
```

src/aiida_pseudo/cli/install.py
```diff
@@ -143,117 +140,215 @@ def download_sssp(
...
 
-        patch_version = download_sssp(configuration, filepath_archive, filepath_metadata, traceback)
+        if from_download is not None:
 
-        description = f'SSSP v{patch_version} {functional} {protocol} installed with aiida-pseudo v{__version__}'
-        description += f'\nArchive pseudos md5: {md5_file(filepath_archive)}'
-        description += f'\nPseudo metadata md5: {md5_file(filepath_metadata)}'
+            tarball_path = pathlib.Path(from_download).absolute()
+            with tarfile.open(tarball_path, 'r') as handle:
+                handle.extractall(dirpath)
```
sphuber, line 216: It is not immediately clear from reading the code that this assumes that the extracted archive will have written the `filepath_archive` and `filepath_metadata` files. Maybe it would help to have an assert, or just a check `if not filepath_archive.exists(): echo.echo_critical('archive `{from_download}` did not contain required file `{filepath_archive.name}`` and same for `filepath_metadata`

sphuber (changes requested):
Thanks @mbercx I think the `cmd_install_sssp` and `cmd_install_pseudo_dojo` commands could do with some internal refactoring to make the logic easier to follow. Only wrote them out for the SSSP one, but same applies to PseudoDojo

mbercx:
I totally missed this review, sorry @sphuber! I'll be picking up this PR soon since it's important for rapid QEapp cloud deployments.

sphuber:
@mbercx I am very much interested in getting this feature in for a use-case we have. Do you have time soon to finalize this? If not, I could also take over if you are ok with that.

mbercx:
> @mbercx I am very much interested in getting this feature in for a use-case we have. Do you have time soon to finalize this? If not, I could also take over if you are ok with that.

I'm moving from EPFL to PSI the next two days, but I could have a look at it over the weekend? Would be good to have it merged for sure.

src/aiida_pseudo/cli/install.py
mbercx: Hmm, I understand how the logic is a bit tricky to follow, but moving the `download_only` up out of the `tempfile` context would lead to some code duplication. In the end the current logic is not too bad, no? Basically, in the `tempfile` context:

```
if from_download is not None:
    # Unpack and check tarball
else: (this is now an extra `if` statement so we only have to check the configuration once)
    # Download archives

if download_only:
    # Pack archives and configuration into tarball
    return

# Install pseudo family
```

sphuber: My point is that you essentially have two scenario's, `scenario A` and `scenario B`, which are completely independent and als mutually exclusive. However, the code currently flows as:
```
scenario A
scenario B
scenario A
```
This is what makes it "difficult" to follow, or "more" difficult then it should be. Why not write it as
```
scenario B
scenario A
```
So just start with the `--download-only` path and return early, and then it is clear that all that remains is for the `--from-download` and normal install path.

src/aiida_pseudo/cli/install.py
```diff
@@ -148,118 +149,252 @@ def download_sssp(
...
+    with attempt('downloading selected pseudopotentials metadata archive... ', include_traceback=traceback):
+        response = requests.get(url_metadata, timeout=30)
+        response.raise_for_status()
+        with open(filepath_metadata, 'wb') as handle:
+            handle.write(response.content)
+            handle.flush()
+
+
+def install_pseudo_dojo(
+    configuration: 'PseudoDojoConfiguration',
```
sphuber, line 324: Think this is not actual used
```suggestion
```

src/aiida_pseudo/cli/install.py
```diff
@@ -297,91 +438,65 @@ def cmd_install_pseudo_dojo(
...
-    paw_configurations = (
-        PseudoDojoConfiguration('1.0', 'PBE', 'SR', 'standard', 'jthxml'),
-        PseudoDojoConfiguration('1.0', 'PBE', 'SR', 'stringent', 'jthxml'),
-        PseudoDojoConfiguration('1.0', 'LDA', 'SR', 'standard', 'jthxml'),
-        PseudoDojoConfiguration('1.0', 'LDA', 'SR', 'stringent', 'jthxml')
-    )
-    # yapf: enable
-
     try:
         pseudo_type = pseudo_type_mapping[pseudo_format]
```
sphuber, line 442: Not introduced by this PR, but this try/except is not necessary since the correct values are already validated by the click option, which is a `Choice`

src/aiida_pseudo/cli/install.py
```diff
@@ -297,91 +438,65 @@ def cmd_install_pseudo_dojo(
...
+            tarball_path = pathlib.Path(from_download).absolute()
+            with tarfile.open(tarball_path, 'r') as handle:
+                handle.extractall(dirpath)
+
+            filepath_configuration = dirpath / 'configuration.json'
+
+            with filepath_configuration.open('r') as handle:
+                configuration = PseudoDojoConfiguration(**json.load(handle))
+                pseudo_type = pseudo_type_mapping[configuration.pseudo_format]
+        else:
```
sphuber, line 470: Here you use an `else` but for the SSSP implementation you do another if after the check that the configuration is valid. Maybe make the two follow the same logic?

src/aiida_pseudo/cli/install.py
```diff
@@ -297,91 +438,65 @@ def cmd_install_pseudo_dojo(
...
-        family.set_default_stringency(default_stringency)
-
-        echo.echo_success(f'installed `{label}` containing {family.count()} pseudopotentials.')
+        description = f'{configuration} installed with aiida-pseudo v{__version__}'
+        description += f'\nArchive pseudos md5: {md5_file(filepath_archive)}'
+        description += f'\nPseudo metadata md5: {md5_file(filepath_metadata)}'
+
+        install_pseudo_dojo(
+            configuration, filepath_archive, filepath_metadata, pseudo_type, label, description, traceback
+        ).set_default_stringency(default_stringency)
```
sphuber, line 502:
```suggestion
        family = install_pseudo_dojo(
            configuration, filepath_archive, filepath_metadata, pseudo_type, label, description, traceback
        )
        family.set_default_stringency(default_stringency)
```

src/aiida_pseudo/cli/install.py
```diff
@@ -148,118 +149,252 @@ def download_sssp(
...
-    :param filepath_archive: absolute filepath to write the pseudopotential archive to.
-    :param filepath_metadata: absolute filepath to write the metadata archive to.
+    :param filepath_archive: path to the pseudopotential archive.
+    :param filepath_metadata: path to the metadata file.
+    :param label: label of the pseudopotential family group.
+    :param description: description of the pseudopotential family group.
     :param traceback: boolean, if true, print the traceback when an exception occurs.
     """
-    from aiida_pseudo.groups.family import PseudoDojoFamily
+    # pylint: disable=too-many-locals
```
sphuber, line 167: does this function really have too many locals?

sphuber (changes requested):

src/aiida_pseudo/cli/install.py
mbercx: For sure, I agree! I only rebased/updated the SSSP so far and pushed the commit since it was getting late and I wanted to continue working on it at PSI. ^^

So the PR wasn't ready for review yet, sorry about that. Typically I'll (re-)request a review once a PR is fully ready.

src/aiida_pseudo/cli/install.py
```diff
@@ -148,118 +149,252 @@ def download_sssp(
...
+    from .utils import attempt, create_family_from_archive
 
-    with attempt('downloading selected pseudopotentials archive... ', include_traceback=traceback):
-        response = requests.get(url_archive, timeout=30)
-        response.raise_for_status()
-        with open(filepath_archive, 'wb') as handle:
-            handle.write(response.content)
-            handle.flush()
+    with open(filepath_metadata, 'rb') as handle:
+        handle.seek(0)
```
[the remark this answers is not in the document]
mbercx: Fair point, I'm not sure why that's there. I simply copied it without thinking about it:

https://github.com/aiidateam/aiida-pseudo/blob/9c7740d95b033e13bc9e691376e86cb5a9913275/src/aiida_pseudo/cli/install.py#L242C1-L244C41

src/aiida_pseudo/cli/install.py
mbercx: Huh, I could've sworn `pylint` was complaining! Must've been late-night delirium.

src/aiida_pseudo/cli/install.py
mbercx: It's used to see if the `configuration` is in the `paw_configurations`, see:

https://github.com/mbercx/aiida-pseudo/blob/3882e06d93884d34643959449b65b642c64f0eb4/src/aiida_pseudo/cli/install.py#L377

mbercx:
@sphuber thanks for the review! I think I addressed all your comments. I've moved the `download_only` logic up, which does lead to some code duplication of:

1. The download call.
2. Creating the `label`.
3. Checking if the `configuration` is valid.

I've also moved the check if the configuration is already installed _before_ the download call in https://github.com/aiidateam/aiida-pseudo/pull/135/commits/3882e06d93884d34643959449b65b642c64f0eb4. This is to avoid pointlessly downloading the pseudos, so the user doesn't see the following:

```
❯ aiida-pseudo install sssp
Report: downloading patch versions information...  [OK]
Report: downloading selected pseudopotentials archive...  [OK]
Report: downloading selected pseudopotentials metadata...  [OK]
Report: SsspFamily<SSSP/1.3/PBE/efficiency> is already installed
```

But instead

```
❯ aiida-pseudo install sssp
Report: SsspFamily<SSSP/1.3/PBE/efficiency> is already installed
```

sphuber (approved):
All good, thanks @mbercx

unkcpz:
@mbercx @sphuber could you make a patch release with this? I need it for the AiiDAlab QE, Thanks.

sphuber:
@unkcpz done: https://pypi.org/project/aiida-pseudo/1.3.0/

unkcpz:
Thanks a lot!

PR #144, opened 2022-12-02
Add support for SSSP v1.2

mbercx:


sphuber:
Any reason this is still a draft? Is anything else needed?

mbercx:
I was still testing it but already wanted to share the update. Seems to be working fine though, so I'll change it to ready for review.

sphuber (approved):
Thanks, all good to go

unkcpz:
Hi both, I have updated the SSSP version to `1.2.1`, I guess there is no need to update this repo since it will read the `versions.yaml` to use the latest patch version, correct?

sphuber:
Indeed, see here:
https://github.com/aiidateam/aiida-pseudo/blob/9ea1570a584c274627d65b97a9f8bc0de788ab11/src/aiida_pseudo/cli/install.py#L116

PR #147, opened 2023-01-24
`RecommendedCutoffMixin`: Remove workaround for stringency units

### LINKED ISSUE #76: `RecommendedCutoffMixin`: remove units workaround for backwards compatibility
In #75 we added a workaround to make sure that all pseudo families installed with `aiida-pseudo>=0.4.0` would keep on functioning as expected after the introduction of units. Although the workaround is not too "nasty", we should remove it at some point in the future and clean up the code.

The tricky part will be to decide when to remove it, since we can never be sure that all users have updated their pseudo families by reinstalling them or using the `get_cutoff_unit` method (directly or indirectly). However, there is no rush in removing the workaround, so maybe we can just remove it in the release of v1.0.0. I've opened this issue so we can keep track of this.

sphuber:
Fixes #76 

A workaround was added in `v0.6` for families with recommended cutoffs that were installed with `v0.5` since that older version didn't explicitly store the units of the stringencies in the extras which the new API assumed would always be there.

The workaround would account for the lack of this extra and would automatically introduce it (in a sense migrating the old family to be compatible with the new API) the first time it would be requested.

Almost two years has passed since the release of `v0.6` which should have been plenty for existing families to have been migrated through this workaround and so it can now be removed.

### DIFF
--- a/src/aiida_pseudo/groups/mixins/cutoffs.py
+++ b/src/aiida_pseudo/groups/mixins/cutoffs.py
@@ -216,7 +216,7 @@ class RecommendedCutoffMixin:
         cutoffs_dict.pop(stringency)
 
         cutoffs_unit_dict = self._get_cutoffs_unit_dict()
-        cutoffs_unit_dict.pop(stringency, None)  # `None` is added to support pseudo families installed with v0.5.0
+        cutoffs_unit_dict.pop(stringency)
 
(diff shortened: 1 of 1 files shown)


mbercx (approved):
Thanks @sphuber, LGTM! 🚀

PR #148, opened 2023-01-24
Release `v1.0.0`

sphuber:


sphuber:
@mbercx I have updated the development classifier in the `pyproject.toml` to `Production/Stable`. Is there anything else that we should change that you can think of?

mbercx (approved):
It's the first time I'm moving a package to v1.0, but there is nothing else I can think of, so good to go for me!

PR #153, opened 2023-05-23
Adding SSSP 1.3.0 support

unkcpz:


unkcpz:
I still need to confirm some detail of the SSSP archive.

sphuber (commented):
Thanks @unkcpz . Do we really need to change the default? And if we change the default version, can we at least keep `efficiency` as the default? What are the changes in SSSP v1.3?

unkcpz:
> What are the changes in SSSP v1.3?

From v1.2 -> v1.3, the pseudopotentials of elements At, Fr, Ra are added from PSlibrary. The pseudopotential of actinides are added from dataset of https://www.uni-marburg.de/de/fb15/arbeitsgruppen/anorganische_chemie/ag-kraus/forschung/paw_datasets_for_the_actinoids. Both efficiency and precision having those pseudos added. Thus I think you are right, it is good to keep the default unchanged.

unkcpz:
But I am confusing on this https://github.com/aiidateam/aiida-pseudo/blob/b331959263efe42bd32f6639464057eb26f8db0f/src/aiida_pseudo/cli/install.py#L112
How this link know which is the latest archive and will find `version.yml` from there?

sphuber:
> But I am confusing on this
> 
> https://github.com/aiidateam/aiida-pseudo/blob/b331959263efe42bd32f6639464057eb26f8db0f/src/aiida_pseudo/cli/install.py#L112
> 
> How this link know which is the latest archive and will find `version.yml` from there?

The link is permanent, and the file `versions.yaml` is updated when a new SSSP version is made available. It is just mapping the minor version to the latest patch. It currently contains:
```
---
# This maps each minor version to the latest patch version
# corresponding to that minor version
'1.0': '1.0'
'1.1': '1.1.2'
'1.2': '1.2.1'
'1.3': '1.3.0'
```
It is just used to go from `1.3` to `1.3.0` in this case.

src/aiida_pseudo/groups/family/sssp.py
```diff
@@ -33,7 +33,7 @@ class SsspFamily(RecommendedCutoffMixin, PseudoPotentialFamily):
 
     label_template = 'SSSP/{version}/{functional}/{protocol}'
     filename_template = 'SSSP_{version}_{functional}_{protocol}'
-    default_configuration = SsspConfiguration('1.1', 'PBE', 'efficiency')
+    default_configuration = SsspConfiguration('1.3', 'PBE', 'precision')
```
unkcpz, line 36:
```suggestion
    default_configuration = SsspConfiguration('1.1', 'PBE', 'efficiency')
```

unkcpz:
It is ready to review, I revert the change of setting default version to latest one and add a comment on the template url.

sphuber:
@unkcpz does the SSSP keep a `CHANGELOG` somewhere? I would like to reference to it to give a summary to users of what changed.

unkcpz:
> @unkcpz does the SSSP keep a CHANGELOG somewhere? I would like to reference to it to give a summary to users of what changed.

Yes it is in https://archive.materialscloud.org/record/file?filename=CHANGELOG.txt&record_id=1732

PR #155, opened 2023-08-22
Add support for Pseudo-dojo v0.5

mbercx:
Add the various configurations/urls for Pseudo-dojo v0.5, and add the option to the CLI install command.

Since the `.djrepo` files are missing the `hints` generated using the delta factor, fall back on the `ppgen_hints` provided by ONCVPSP in this case.


sphuber (approved):
Code looks good. If it actually works in the wild, fine to merge

PR #157, opened 2023-08-30
Backport of support for SSSP v1.3 and PseudoDojo v0.5

sphuber:


mbercx (approved):
Did some field testing and the new versions seem to install fine!

Also seems that the latest `docutils` doesn't play nice with our quite old Sphinx version, so I added an upper limit to remove the warning.

PR #160, opened 2023-09-12
Docs: Improve section on installing custom archive

sphuber:
Add more information on how the command will fail if the pseudopotential files do not respect the expected naming scheme and how the format can be specified explicitly as a fix.

I think with these instructions, the `aiida.orm.nodes.data.upf` module can be deprecated as in principle any family can be created.


sphuber:
Appreciate the heart @mbercx but does anything need to be changed here, or can we merge it?

PR #161, opened 2023-09-22
Docs: Add how to migrate from legacy UPF family

sphuber:
The provided snippet can be used to migrate an existing UPF family, created with the `aiida.orm.nodes.data.upf` module from `aiida-core`. A `PseudoPotentialFamily` will be created to which the existing `UpfData` nodes are added, after having been converted to the `UpfData` class of `aiida-pseudo`.

### DIFF
--- a/docs/source/howto.rst
+++ b/docs/source/howto.rst
@@ -136,5 +136,52 @@ where ``<STRINGENCY>`` is a string that defines the recommended cutoffs, ``<FAMI
     The ``PseudoPotentialFamily`` base family type does not support setting recommended cutoffs.
     To be able to use this feature for a manually installed family, install it as a ``CutoffsPseudoPotentialFamily`` as expained :ref:`here <how-to:install_archive:family_type>`.
 
+
+.. _how-to:install_archive:migrate-legacy:
+
+Migrate from legacy ``UpfData`` from ``aiida-core``
+---------------------------------------------------
+
+Since the earliest versions, ``aiida-core`` provided its own data plugin to support UPF pseudopotentials as well as a concept of a pseudopotential family.
+The ``UpfData`` from ``aiida-core`` and related utilities are now deprecated and are replaced by ``aiida-pseudo``.
+To migrate an existing UPF family from ``aiida-core``, use the following snippet:
+
+.. code-block:: python
+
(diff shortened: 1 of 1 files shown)


sphuber:
@mbercx @mkotiuga any feedback on this? Otherwise I am going to merge this soon

docs/source/howto.rst
```diff
@@ -128,5 +128,52 @@ where ``<STRINGENCY>`` is a string that defines the recommended cutoffs, ``<FAMI
...
+    legacy_family = load_group(label)
+    family = PseudoPotentialFamily(label=label_new).store()
+    pseudos = []
+
+    for legacy_pseudo in legacy_family.nodes:
+        with legacy_pseudo.open(mode='rb') as handle:
+            pseudos.append(UpfData(handle, filename=legacy_pseudo.filename).store())
+
+    family.add_nodes(pseudos)
+    print(f'Migrated `{legacy_family}` to `{family}`.')
```
mkotiuga, line 159: If cutoffs also need to be set, maybe prompt the user here?

sphuber: Cutoffs are not required. And since legacy families don't even have this concept, there is nothing to migrate really. Another part of the docs that explains how to manually create/install families goes over how to add cutoffs

mkotiuga: Are you sure they are not required? did this change? I remember I could not use the pseudofamilies I had made myself until I had set cutoffs for all the pseudos

sphuber: The only thing that won't work is the `PwBaseWorkChain.get_builder_from_protocol` because that method will want to automatically set the cutoffs, and so if the family doesn't define these, it raises. However, nothing is stopping a user from defining the inputs manually. The family is just a `Group` with convenience methods such as `PseudoFamily.get_pseudos(structure=structure)` to automatically get the dictionary of pseudos that can be passed to the `pseudos` input.

We could perhaps think of making the cutoffs optional in the `get_builder_from_protocol` and not have it raise, but simply warn, so the user can specify them manually directly in the `parameters` input. But this is a feature request for `aiida-quantumespresso` really, and has nothing to do with `aiida-pseudo`.

PR #166, opened 2023-10-24
CLI: Enable install `--download-only` without available profile

### LINKED ISSUE #165: Install with --download-only require a valid AiiDA profile
In preparing the AiiDAlab QE app demo image, I tried to use `--download-only` to download pseudos first and then install the pseudos from the file which will make the first-time app start faster. 
However, I need to call install with `--download-only` in an environment without aiida profile and it raises an error. 
I notice the `options.PROFILE()` is set for the command root thus every sub command needs it as well. Is there a way that when `--download-only` is set the PROFILE option be excluded temporarily?

sphuber:
Fixes #165 

The `aiida-pseudo install sssp/pseudo-dojo` provide the `--download-only` option to only download the archive. This is useful in case a family needs to be installed on a machine without a (stable) internet connection. With this option, the archive can be downloaded on another machine first before transferring the file and installing it directly from the archive.

Currently, however, the install command always requires an AiiDA profile to be configured, even for the `--download-only` option in which case the command never needs access to the profile. Here, this requirement is removed by removing the `with_dbenv` decorator from the command. Instead, the profile is loaded manually inside the command at the point where it is really necessary.

### DIFF
--- a/src/aiida_pseudo/cli/install.py
+++ b/src/aiida_pseudo/cli/install.py
@@ -198,7 +198,6 @@ def install_sssp(
 @options.DOWNLOAD_ONLY()
 @options.FROM_DOWNLOAD()
 @options.TRACEBACK()
-@decorators.with_dbenv()
 def cmd_install_sssp(version, functional, protocol, download_only, from_download, traceback):
     """Install an SSSP configuration.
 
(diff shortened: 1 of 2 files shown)


sphuber:
@unkcpz I think this actually works. Could you give this a go please?

unkcpz:
Wow, this is lightning-fast! Thanks, I'll test it.

unkcpz:
Tested and works!!

unkcpz (approved):
The changes look all good and thanks for adding the test.

Here is how I use it in QeApp docker image,
https://github.com/aiidalab/aiidalab-qe/blob/e1c57376840ec2686e49ce09439f1d653d6c6587/docker/Dockerfile#L34-L40
Install from pre-downloaded files saves ~3s for each pseudo library with a test on my workstation at PSI. Could be better for servers with worse network conditions. 

Can you give it a patch release after this is merged?

PR #168, opened 2023-12-08
Hashing made with element and md5 checksum

### LINKED ISSUE #167: Ignore version and filename when compute hash for the pseudo node
A typical UpfData node has the following objects used to generate the hash:

```
['1.4.0',                                                                                                                                    
 {'md5': 'e96aa7e3e4a958db16958554ea960e2a',                                                                                                     
  'element': 'H',                                                                                                                                
  'filename': 'H.nc.z_1.oncvpsp3.dojo.v0.5.0-std.upf',                                                                                       
  'z_valence': 1},                                                      
 '2ad1f9704b3b12e280562806303ce715cbd9f4735c24b047199cbbcd47e168e8',    
 None] 
```

I think the filename and the version of the plugin are relevant to the calculation result thus need to add to `_hash_ignored_attributes`. Your opinion @sphuber ?

unkcpz:
fixes #167 

For pseudo data, all we care about for the node is the hash of the file. So we could even think of overriding the _get_objects_to_hash and just return the md5 and element.

### DIFF
--- a/src/aiida_pseudo/data/pseudo/pseudo.py
+++ b/src/aiida_pseudo/data/pseudo/pseudo.py
@@ -1,4 +1,6 @@
 """Base class for data types representing pseudo potentials."""
+from __future__ import annotations
+
 import io
 import pathlib
 import typing
@@ -7,8 +9,17 @@ from aiida import orm, plugins
 from aiida.common.constants import elements
 from aiida.common.exceptions import StoringNotAllowed
 from aiida.common.files import md5_from_filelike
+from aiida.orm.nodes.caching import NodeCaching
 
 __all__ = ('PseudoPotentialData',)
+
+
+class PseudoPotentialDataCaching(NodeCaching):
+    """Class to define caching behavior of ``PseudoPotentialData`` nodes."""
+
+    def _get_objects_to_hash(self) -> list:
+        """Return a list of objects which should be included in the node hash."""
+        return [self._node.element, self._node.md5]
 
 
 class PseudoPotentialData(plugins.DataFactory('core.singlefile')):
@@ -16,6 +27,8 @@ class PseudoPotentialData(plugins.DataFactory('core.singlefile')):
 
     _key_element = 'element'
     _key_md5 = 'md5'
+
+    _CLS_NODE_CACHING = PseudoPotentialDataCaching
 
     @classmethod
     def get_or_create(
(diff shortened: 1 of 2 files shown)


unkcpz:
Somehow the objcets are the same but the hash values are different? I think I miss some key behaviour of `make_hash` function.

sphuber:
It is not a misunderstanding of the `make_hash` function but rather that `Node.get_hash` does _not_ call `Node._get_objects_to_hash()` but `Node.base.caching._get_objects_to_hash`. You should override the `Node._CLS_NODE_CACHING` attribute with a custom version of `NodeCaching`, i.e., something like:
```python
from aiida.orm.nodes.caching import NodeCaching

class PseudoPotentialDataCaching(NodeCaching):
    """Class to define caching behavior of ``PseudoPotentialData`` nodes."""

    def _get_objects_to_hash(self) -> list:
        """Return a list of objects which should be included in the node hash."""
        return [self.element, self.md5]


class PseudoPotentialData(Data):

    _CLS_NODE_CACHING = PseudoPotentialDataCaching
```
That should make the tests pass. You have to change the `original._get_objects_to_hash` to `original.base.caching._get_objects_to_hash` of course.

unkcpz:
@sphuber It works! thanks!

sphuber:
Last thing I was wondering @unkcpz . Why do we need to include the element? Should the file content not really just be sufficient? Could it really be possible where you have two identical files but they should be for different elements?

unkcpz:
> Could it really be possible where you have two identical files but they should be for different elements?

Because I saw the setter function for the element and then I was thinking that there might be different site labels for the same element using the same UPF file. 
I think it all depends on whether we assume user will change the element attribute or not. Afterall, the element is parsed from file, the same md5 will guarantee the same element.

sphuber:
> Afterall, the element is parsed from file, the same md5 will guarantee the same element.

For the subclasses it is yes, but for the `PseudoPotentialData` it has to be set manually. But regardless on how it is set, if the file content is identical for two different pseudos, how could it _possibly_ for two differente elements? Functionally, there would be zero difference, so would it be necessary to even include the element attribute?

unkcpz:
I don't have any use case involved with subfix for element, so I am leaning to remove element from caching.
Is there use case in aiida-quantumespresso that may set for instance `Fe` to `Fe_1` and `Fe_2`, even so that should have nothing to do with the aiida-pseudo I guess?

sphuber (approved):
Thanks @unkcpz . Let's get this merged. The concept just makes sense. We will find out later if there really is some use-case out there that is not compatible with this change.

PR #173, opened 2024-10-28
Udjusted the documentation regarding the creating of a pseudo family.

Minotakm:
Hey All, 

I tried to use the documentation available regarding the design of a new pseudo-family and I found I need to make a few small-ish adjustments to make it work. I thought it would be nice to add them in the documentation. This pull request includes updates to the code snippets on the design section of the documentation, reflecting the adjustments I made. Looking forward to your feedback.

### DIFF
--- a/docs/source/design.rst
+++ b/docs/source/design.rst
@@ -71,17 +71,17 @@ A pseudopotential family can be constructed manually, by first constructing the 
     from aiida import plugins
 
     UpfData = plugins.DataFactory('pseudo.upf')
-    PseudoPotentialFamily = plugins.GroupFactory('pseudo')
+    PseudoPotentialFamily = plugins.GroupFactory(entry_point_name='pseudo.family')
 
     pseudos = []
 
     for filepath in ['Ga.upf', 'As.upf']:
         with open(filepath, 'rb') as stream:
-            pseudo = UpfData(stream)
-            pseudos.append(pseudo)
+            pseudo = UpfData(file=stream)
+            pseudos.append(pseudo.store())
 
     family = PseudoPotentialFamily(label='pseudos/upf').store()
-    family.append(pseudos)
+    family.add_nodes(pseudos)
 
 Note that as with any :py:class:`~aiida.orm.Group`, it has to be stored before nodes can be added.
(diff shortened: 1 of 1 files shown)


docs/source/design.rst
```diff
@@ -71,17 +71,17 @@ A pseudopotential family can be constructed manually, by first constructing the
     from aiida import plugins
 
     UpfData = plugins.DataFactory('pseudo.upf')
-    PseudoPotentialFamily = plugins.GroupFactory('pseudo')
+    PseudoPotentialFamily = plugins.GroupFactory(entry_point_name='pseudo.family')
```
sphuber, line 74: Don't think this change is necessary, is it?
```suggestion
    PseudoPotentialFamily = plugins.GroupFactory('pseudo.family')
```

docs/source/design.rst
```diff
@@ -71,17 +71,17 @@ A pseudopotential family can be constructed manually, by first constructing the
...
 
     pseudos = []
 
     for filepath in ['Ga.upf', 'As.upf']:
         with open(filepath, 'rb') as stream:
-            pseudo = UpfData(stream)
-            pseudos.append(pseudo)
+            pseudo = UpfData(file=stream)
```
sphuber, line 80: This isn't necessary either I am pretty sure
```suggestion
            pseudo = UpfData(stream)
```

sphuber (changes requested):
Thanks a lot for the effort @Minotakm . You did indeed catch a number of mistakes. There were just two changes that are not necessary I believe, and if true, I would prefer to leave it as is. Let me know if reverting those two small changes still leads to a problem on your end.

docs/source/design.rst
Minotakm: Yeah, I just put it to remember what it is, we can leave it as it is.

docs/source/design.rst
Minotakm: The same here. Although, I forgot to add the .store() method in the pseudo, which without it creates an error. The working example should be: pseudos.append(pseudo.store())

Minotakm: Should be fine now. Thanks

sphuber:
Will merge as soon as I fix the current broken CI. https://github.com/aiidateam/aiida-pseudo/pull/175

PR #179, opened 2024-12-20
Proposed fix for HTTPSConnection Issue with PseudoDojo

Minotakm:
Hey  All, 

I tried to download a version of pseudo dojo pseudopotentials however I couldn't due to : 

```
Report: downloading selected pseudopotentials archive...  [FAILED]
Critical: HTTPSConnectionPool(host='www.pseudo-dojo.org', port=443): Max retries exceeded with url: /pseudos/nc-fr-04_pbesol_standard_psp8.tgz (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1007)')))
```

As a quick fix we can bypass the certificate check when we download the specific pseudopotentials. Happy to hear any feedback. Thanks :D

PR #181, opened 2024-12-23
Move `aiida-core~=2.6` dependency to dev

### LINKED ISSUE #180: Dependency conflict
a73028fc2a48938dffe92ed9abff0207f818bc0c changed the `aiida-core` dependency to `~=2.6` for the purpose of leveraging `aiida-core` testing fixtures. However, this dependency leads to conflicts with packages using `aiida-pseudo` that have yet to upgrade.

edan-bainglass:
This PR reverts the `aiida-core` dependency back to `~=2.1` and moves the `~=2.6` pin to `dev` dependencies. See resolved issue for detail.

Resolves #180

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,7 @@ classifiers = [
   'Programming Language :: Python :: 3.12'
 ]
 dependencies = [
-  'aiida-core~=2.6',
+  'aiida-core~=2.1',
   'click~=8.0',
   'pint~=0.23.0',
   'requests~=2.20'
@@ -46,6 +46,7 @@ requires-python = '>=3.9'
 
 [project.optional-dependencies]
 dev = [
+  'aiida-core~=2.6',
   'pre-commit~=2.2',
   'pytest>=6.0'
 ]


sphuber:
Unfortunately it is not so easy. There were a number of changes in the API of `aiida-core` in v2.6 that we already adjusted to, amongst which the hashing. See bcc0129cc0b1b6fa5ddfe7d91f09d805ceb40d15

The issue says

> However, this dependency leads to conflicts with packages using aiida-pseudo that have yet to upgrade.

What are the specific problems? I think we should rather work to fix those than revert here.

edan-bainglass:
> The issue says
> 
> > However, this dependency leads to conflicts with packages using aiida-pseudo that have yet to upgrade.
> 
> What are the specific problems? I think we should rather work to fix those than revert here.

I'll let @superstar54 explain. It was mostly a concern for those who use older versions of `aiida-core` AND `aiida-pseudo`.

As for the particular issue at hand, we're trying to release `aiidalab-qe`. All of our tests are running into an issue with `aiida-pseudo` w.r.t downloading the pseudos. This was "fixed" in #179, which already includes the `~=2.6` pin. This is now causing issues with the release due to `aiidalab` dependencies, but let me see if I can overcome those issues. I think it has already been tested with 2.6 and a new image is available. I'll check.

superstar54:
Hi @sphuber , thanks for the quick comment. 

> What are the specific problems? I think we should rather work to fix those than revert here.

We want to use `aiida-pseudo` inside the AiiDAlab docker image, which fixs `aiida-core` version==2.5.1.

superstar54:
We avoid the issue by using the latest AiiDAlab full-stack image with aiida-core 2.6.

As discussed with @edan-bainglass , we closed this PR.

PR #185, opened 2025-04-18
update pint dependency to python 3.13

rikigigi:
Hello,

a simple upgrade of a dependency does not crash aiida when importing it in python 3.13

https://github.com/hgrecco/pint/issues/2065

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ classifiers = [
 dependencies = [
   'aiida-core~=2.6',
   'click~=8.0,<8.2',
-  'pint~=0.23.0',
+  'pint>=0.23.0,<0.25.0',
   'requests~=2.20'
 ]
 dynamic = ['description', 'version']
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -19,7 +19,7 @@ classifiers = [
 dependencies = [
   'aiida-core~=2.6',
   'click~=8.0',
-  'pint~=0.23.0',
+  'pint~=0.24.4',
```
sphuber, line 22:
```suggestion
  'pint>=0.23.0,<0.25.0',
```

sphuber (changes requested):
Thanks for the contribution @rikigigi . Sorry for the late review. Instead of dropping Python 3.9 support just to accommodate this change, why not just loosen the range. This way we can keep support for 3.9 until we remove that ourselves. Prefer not to merge these kinds of changes

sphuber (approved):
Thanks @rikigigi , sorry for the delay. Will make a release soon

PR #186, opened 2025-05-19
Spruce up CI

danielhollas:
These tests were still living in the aiida-quantumespresso repository so moving them here instead.

Incidentally, these tests started failing, so we need to figure out why they are failing.

EDIT: The failing tests actually come from the new release of click (8.2.0), which apparently broke a lot of people.

https://github.com/pallets/click/issues/2908

For now pinning click does the trick, I suspect the actual issue needs to be fixed in aiida-core?

I've also updated CI config and added uv for much faster installation.


.github/workflows/ci.yml
```diff
@@ -2,51 +2,63 @@ name: ci
...
+        activate-environment: true
 
     - name: Install Python package and dependencies
-      run: pip install -e .[dev]
+      run: uv pip install -e .[dev]
 
     - name: Run pytest
       env:
         AIIDA_WARN_v3: true
-      run: pytest -sv tests
```
danielhollas, line 52: This is extremely verbose and makes it hard to see errors.

.pre-commit-config.yaml
```diff
@@ -12,11 +12,6 @@ repos:
     args: [--fix=lf]
   - id: trailing-whitespace
 
-- repo: https://github.com/ikamensh/flynt/
```
danielhollas, line 15: This is now part of ruff (for the most part) so not needed.

tests/cli/test_commands.py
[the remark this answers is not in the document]
sphuber: Thanks @danielhollas , but this move is not necessary. The tests are already here: https://github.com/aiidateam/aiida-pseudo/blob/main/tests/cli/test_root.py
It was just a typo in `aiida-quantumespresso`. The tests should remain there but just correct the import.

sphuber (changes requested):
Thanks for the PR @danielhollas . As mentioned in the comment, please remove the added tests.

tests/cli/test_commands.py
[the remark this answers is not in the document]
danielhollas: Reverted, thanks!

sphuber:
> I think the easiest thing here would be to not depend on the hugely complex verdi cli, we can vendor the bits we need here. Marking the PR as draft for now.

Well, the CLI actually relies on a number of functionalities provided by the implementation of `aiida-core`, such as support for profiles and other things. Don't think it is wise to start reimplementing this here, especially since `aiida-core` is a dependency anyway and so its CLI basics will always be there.

.github/workflows/ci.yml
```diff
@@ -2,51 +2,63 @@ name: ci
...
+    - uses: actions/checkout@v4
 
     - name: Install Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v4
+      uses: actions/setup-python@v5
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - name: Install uv
+      uses: astral-sh/setup-uv@v6
```
sphuber, line 53: Why are you switching to uv in the CI but not the CD? I think it is good to keep the same approach for both, to minimize failures of CD even though CI is passing

danielhollas: I added uv there as well, I didn't realize you're running tests in cd.yml.

danielhollas:
> Well, the CLI actually relies on a number of functionalities provided by the implementation of aiida-core, such as support for profiles and other things. Don't think it is wise to start reimplementing this here, especially since aiida-core is a dependency anyway and so its CLI basics will always be there.

Yes, I realized quite quickly how ingrained verdi stuff is in here, so getting rid of it is not viable. Let's go with this PR for now. I have an idea how to fix click, but it would be a breaking change for the `aiida-pseudo list` command so I'll do that in a separate PR where we can discuss.

danielhollas:
Haha, you were faster and already merged the pin :sweat_smile: I've renamed the title of this PR accordingly then...

.github/workflows/ci.yml
sphuber: Really need to look sometime whether it is possible to use templates and get rid of the duplication. Is annoying to have to maintain both CI and CD workflows that are so similar

sphuber (approved):
thanks @danielhollas

PR #190, opened 2025-06-30
♻️ Fix tests

mbercx:
The `empty_config` context manager is used for some of the testing infrastructure of the installation commands. It relied on some methods for adapting the AiiDA configuration folder, whose API has been changed in

https://github.com/aiidateam/aiida-core/commit/9baf3ca96caa5577ec8ed6cef69512f430bb5675

Here we adapt the `empty_config` context manager to rely on the new API using the `AiiDAConfigDir`.

### DIFF
--- a/tests/data/pseudo/test_pseudo.py
+++ b/tests/data/pseudo/test_pseudo.py
@@ -32,12 +32,6 @@ def test_constructor_source_types(source):
     assert not pseudo.is_stored
 
 
-def test_constructor_invalid():
-    """Test the constructor for invalid arguments."""
-    with pytest.raises(TypeError, match='missing 1 required positional argument'):
-        PseudoPotentialData()
-
-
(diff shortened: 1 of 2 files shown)


tests/data/pseudo/test_pseudo.py
```diff
@@ -32,12 +32,6 @@ def test_constructor_source_types(source):
     assert not pseudo.is_stored
 
 
-def test_constructor_invalid():
-    """Test the constructor for invalid arguments."""
-    with pytest.raises(TypeError, match='missing 1 required positional argument'):
-        PseudoPotentialData()
-
-
```
edan-bainglass, line 40: No longer needed?

edan-bainglass: Never mind. You explain it clearly in the commit message 👍

edan-bainglass (approved):
LGTM!

PR #193, opened 2025-08-04
Optionally allow `None` for missing pseudos in a family

edan-bainglass:
At the moment, using `PseudoPotentialFamily.get_pseudos()` will fail for all if at least one element does not have a corresponding pseudopotential in the family. This PR aims to allow a path for partial results, returning `None` for those missing elements, if user opt in by way of an optional `none_if_missing` parameter (`False` by default for backwards compatibility).

This allows the QE app, for example, to do this:

image

### DIFF
--- a/src/aiida_pseudo/groups/family/pseudo.py
+++ b/src/aiida_pseudo/groups/family/pseudo.py
@@ -347,11 +347,7 @@ class PseudoPotentialFamily(Group):
 
         if structure is not None:
             return {
-                kind.name: self.get_pseudo(kind.symbol, none_if_missing=none_if_missing)
-                for kind in structure.kinds
+                kind.name: self.get_pseudo(kind.symbol, none_if_missing=none_if_missing) for kind in structure.kinds
             }
 
-        return {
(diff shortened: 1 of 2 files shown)


edan-bainglass:
@t-reents @mbercx no rush on this and also okay to drop. I realized I'm the app I can just replace get_pseudos() with a loop over get_pseudo() and try/except missing pseudos, replacing them with None. So this API is not strictly required.

I'll let you decide if this should close or merge. I'm good either way.

t-reents:
True, that's a solution as well!

As I told you in person, I was thinking that this change is fine, as it's backwards compatible and having in mind the use-case in the app. 
In my opinion, the use-case in the app is/was the strongest motivation for this change. For a normal user, I still feel that it makes sense to raise in case a pseudo is missing, since one is explicitly trying to get the pseudos for a given structure/set of elements from a `PseudoFamily` (even though I also see the other side to provide at least the available pseudos).

In summary, no strong opinion at my end but a preference towards the current behavior. I'd leave it up to @mbercx to make a final decision

PR #201, opened 2025-12-05
Add `--all` option to install all SSSP or PseudoDojo functionals/protocols

### LINKED ISSUE #101: CLI: Add `--all` option to automated install commands
Since some users might simply want install all pseudopotential configurations of a certain family (I'm running into this when installing pseudos for the [QEapp](https://github.com/aiidalab/aiidalab-qe) for example), it would be good to add an `--all` options that simply installs all configurations.

t-reents:
This new option is something that we already discussed several times. It allows the user to directly install all combinations of protocols and functionals for SSSP and PseudoDojo for the specified version etc.

Pinging @mbercx @Minotakm @giovannipizzi @mikibonacci

### DIFF
--- a/src/aiida_pseudo/cli/install.py
+++ b/src/aiida_pseudo/cli/install.py
@@ -202,8 +202,18 @@ def install_sssp(
 @options.PROTOCOL(type=click.Choice(['efficiency', 'precision']), default='efficiency', show_default=True)
 @options.DOWNLOAD_ONLY()
 @options.FROM_DOWNLOAD()
+@click.option(
+    '--all',
+    'install_all',
+    is_flag=True,
+    default=False,
+    help=(
+        'Install all SSSP functionals and protocols (PBE and PBEsol for the efficiency and '
+        'precision protocols) for the specified version.'
+    ),
+)
 @options.TRACEBACK()
-def cmd_install_sssp(version, functional, protocol, download_only, from_download, traceback):
+def cmd_install_sssp(version, functional, protocol, download_only, from_download, install_all, traceback):
     """Install an SSSP configuration.
 
     The SSSP configuration will be automatically downloaded from the Materials Cloud Archive entry to create a new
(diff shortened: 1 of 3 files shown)


giovannipizzi:
Thanks! Since people are lazy, would it be OK to fins a way toake this the default, and turn this off if one specifies a given version? What do you think?

src/aiida_pseudo/cli/install.py
```diff
@@ -461,70 +506,88 @@ def cmd_install_pseudo_dojo(
     }
     pseudo_type = pseudo_type_mapping[pseudo_format]
 
-    configuration = PseudoDojoConfiguration(version, functional, relativistic, protocol, pseudo_format)
-    label = PseudoDojoFamily.format_configuration_label(configuration)
+    configurations = [(version, functional, relativistic, protocol, pseudo_format)]
+    if install_all:
+        configurations = [
+            (version, 'PBE', relativistic, 'standard', pseudo_format),
```
mbercx, line 512: A note from the [PseudoDojo website](https://www.pseudo-dojo.org/):

> Warning! In the v0.5, the following elements have been updated: Ba, Bi, I, Pb, Po, Rb, Rn, S, Te, Tl, and Xe. They are only available for PBE. Use the version 0.4.1 for LDA or PBEsol xc.

So the "latest" version is functional-dependent here.

mbercx:
> Since people are lazy, would it be OK to fins a way toake this the default

Hmm, have to think about this. Initially I was against it, but it might be the most natural. Alternatively (and the reason I came here to comment), is to call the option `--latest`, because that is what it really installs, not `--all`. But perhaps `--latest` doesn't convey that you are installing multiple versions.

However, for PseudoDojo this is still not entirely true, because you install the latest version for all functionals and "precision levels"/"protocols". Not all formats. This is a sensible decision, since otherwise you'd be installing 24 families. That would take a while, especiall with the current connection issues. ^^ The fact that here you really have to select the correct format for the pseudos to be useful, and that you install 6 spurious families in case you select the wrong one, makes me apprehensive towards having `--all` as the default behaviour.

